PR #215, opened 2024-06-10
FEAT Set log level from config

GeorgWa:
Allows setting log level from config.
Please review #216 first.


alphadia/workflow/reporting.py
```diff
@@ -25,7 +25,7 @@
 
 # Add a new logging level to the default logger
 # This has to happen at load time to make the .progress() method available even if no logger is instantiated
-PROGRESS_LEVELV_NUM = 100
+PROGRESS_LEVELV_NUM = 21
```
mschwoer, line 28: mb comment what the 21 means?

alphadia/planning.py
```diff
@@ -123,6 +123,14 @@ def __init__(
         if "output" not in self.config:
             self.config["output"] = output_folder
 
+        # set log level
+        level_to_set = self.config["general"]["log_level"]
+        level_code = logging.getLevelNamesMapping().get(level_to_set)
+        if level_code is None:
```
mschwoer, line 129: `if (level_code:=logging.getLevelNamesMapping().get(level_to_set)) is None:`

mschwoer (approved):
Logs good to me!

PR #218, opened 2024-06-10
FIX transfer learning bug

GeorgWa:
This PR fixes two bugs during transfer learning:
- I modified a method argument in place which is a global variable in python
- Decoys are now removed from the transfer learning dataset


mschwoer (approved):
LGTM

alphadia/workflow/peptidecentric.py
```diff
@@ -1160,13 +1164,13 @@ def _build_candidate_speclib_flat(
         Dataframe with scored candidates
     """
     # remove decoys
-    psm_df = psm_df[psm_df["decoy"] == 0]
+    # psm_df = psm_df[psm_df["decoy"] == 0]
 
-    for col in ["rt_observed", "mobility_observed", "mz_observed"]:
-        optional_columns += [col] if col in psm_df.columns else []
+    # make copy to avoid modifying the original dataframe
```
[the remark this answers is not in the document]
mschwoer: now

mschwoer, line 1169: please remove the mutable arguments..

https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

I guess that once we removed the "source of all evil", life will finally be great on planet earth .. so this seems to be a unique chance! :-)

mschwoer (changes requested):
please remove mutable arguments where they caused issues

alphadia/workflow/peptidecentric.py
GeorgWa: I used the pattern from the blog,  it's really a dangerous pattern

alphadia/workflow/peptidecentric.py
```diff
@@ -1102,22 +1102,7 @@ def _build_candidate_speclib_flat(
...
-        "mobility_library",
-        "genes",
-        "proteins",
-        "decoy",
-        "mods",
-        "mod_sites",
-        "sequence",
-        "charge",
-    ],
+    optional_columns: typing.Union[typing.List[str], None] = None,
```
mschwoer, line 1105: (nit) we should move the whole codebase from `typing.Union[typing.List[str], None]` to ` list | None`. (not in this PR :-))

mschwoer (approved):
LGTM

alphadia/workflow/peptidecentric.py
mschwoer: @jalew188 FYI ;-)

alphadia/workflow/peptidecentric.py
GeorgWa: hahaha, what's the idea behind this? I guess my typing knowledge is deprecated :D

PR #223, opened 2024-06-11
FEAT automatic transfer learning in consensus pipeline

GeorgWa:
Allows automatic transfer learning of RT, charge, MS2 as part of the regular search workflow.
Introduces a new config group to the `.yaml`.

```yaml
transfer_learning:

  # if true, a custom peptdeep model will be created using the transfer learned library
  enabled: False

  # number of precursors per batch
  batch_size: 2000

  # maximum learning rate per batch.
  # The maximum learning rate will be reached after a warmup phase and decreased using a plateau scheduler
  max_lr: 0.0001

  # TODO remove and replaced by fixed 70:20:10 split
  train_ratio: 0.8

  # test every n intervals
  test_interval: 1

  # learning rate patience after which the lr will be halved
  lr_patience: 3

  # minimum precursor number to perform transfer learning
  minimum_psms: 10000

  # maximum number of epochs
  epochs: 51

  # number of warmup epochs during which the lr is ramped up
  warmup_epochs: 5

  # normalised collision energy encoded during training
  nce: 25

  # instrument type encoded during training
  instrument: 'Lumos'
  ```

### CODE REVIEW

File: alphadia/constants/default.yaml [hunk 1]
```diff
@@ -141,8 +141,8 @@ optimization_manager:
 
 # This section controls transfer learning
```
mschwoer, line 142: the `default.yaml` in `misc/config` can be deleted I guess?

alphadia/outputtransform.py
```diff
@@ -374,9 +376,44 @@ def build(
...
+    def build_transfer_model(self):
+        logger.progress("Train PeptDeep Models")
+
+        transfer_lib_path = os.path.join(
+            self.output_folder, f"{self.TRANSFER_OUTPUT}.hdf"
+        )
+        if not os.path.exists:
+            raise ValueError(
+                f"Transfer learning library was not found at {self.TRANSFER_OUTPUT}.hdf. Did you enable library generation?"
+            )
```
mschwoer, line 394: I saw this pattern a log today already ..
Think about introducing one static method that does this check
```
def check_path_exists(path, "Transfer learning library was not found. Did you enable library generation?")
... # works for files and folders
```

alphadia/outputtransform.py
```diff
@@ -374,9 +376,44 @@ def build(
...
+
+        transfer_lib = SpecLibBase()
+        transfer_lib.load_hdf(
+            transfer_lib_path,
+            load_mod_seq=True,
+        )
+
+        device = "cpu"
+        if self.config["general"]["use_gpu"]:
+            device = "mps" if os.uname().sysname == "Darwin" else "gpu"
```
mschwoer, line 404: here comes the `uname` bug again .. https://github.com/MannLabs/alphadia/pull/201 :-)
please create a static method that does this decision and call it here and in `libtransform.py`

alphadia/constants/default.yaml
```diff
@@ -141,8 +141,8 @@ optimization_manager:
...
 # currently only the library is created with transfer learning
-transfer_learning:
-  # if true, the library is created with transfer learning
+transfer_library:
```
mschwoer, line 144: (very) strictly speaking, this a a breaking change

image

mschwoer (changes requested):
please centralize the code around `uname` and remove the bug

alphadia/constants/default.yaml
GeorgWa: I could settle for 1.7.0?

mschwoer: fair enough, I guess there will be a lot of these "strictly speaking breaking changes" in the near future, and we don't want to blow up the major version too much
(writing this using Firefox 126.0.1 btw :-D )

alphadia/outputtransform.py
GeorgWa: image

alphadia/outputtransform.py
GeorgWa: I've added an assert one-liner, what do you think?

alphadia/libtransform.py
```diff
@@ -298,14 +298,7 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
 
-        # Check if CPU or GPU/MPS should be used
-        device = "cpu"
-        if self.use_gpu:
-            try:
-                device = "mps" if os.uname().sysname == "Darwin" else "gpu"
-            except AttributeError:
-                # Windows does not support uname
-                device = "gpu"
+        device = utils.get_torch_device(self.use_gpu)
```
jalew188, line 301: if os is Darwin, "mps" is also possibly not available.

GeorgWa: Very true, I thought this might be handled by peptdeep but I've included it now.

alphadia/outputtransform.py
```diff
@@ -374,9 +376,41 @@ def build(
...
+        if self.config["transfer_library"]["enabled"]:
             _ = self.build_transfer_library(folder_list, save=True)
 
+        if self.config["transfer_learning"]["enabled"]:
+            _ = self.build_transfer_model()
+
+    def build_transfer_model(self):
+        logger.progress("Train PeptDeep Models")
+
+        transfer_lib_path = os.path.join(
```
mschwoer, line 388: All these path check could be moved to one central place that is run at the very beginning .. (new ticket)

mschwoer (approved):
LGTM!

PR #229, opened 2024-06-15
FEAT modular FDR core implementation

GeorgWa:
This is the core FDR implementation for more modular FDR calculation across alphaDIA.
This PR introduces the fdrx module (to be renamed) which consolidates `fdr.py` and `fdrexperimental.py`.


alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
+# make it an abc
```
mschwoer, line 1: please add a sensible module docstring :-)

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+logger = logging.getLogger()
+
+
+class TargetDecoyFDR:
+    def __init__(
+        self,
+        classifier: sklearn.base.BaseEstimator,
+        feature_columns: list,
+        decoy_column: str = "decoy",
+        competition_columns: list = [],
```
mschwoer, line 26: mutable argument

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+        X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(
+            X, y, test_size=0.2
+        )
+        self._classifier.fit(X_train, y_train)
+
+        # evaluate classifier
+        y_test_proba = self._classifier.predict_proba(X_test)[:, 1]
+        y_train_proba = self._classifier.predict_proba(X_train)[:, 1]
+
+        _plot_score_distribution(y_train, y_train_proba, y_test, y_test_proba)
```
mschwoer, line 78: hm, why is this plotting? isn't this supposed to be called during a workflow? where would the plots go then?

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+        is_na_row = psm_df[self._feature_columns].isna().any(axis=1)
+        X = psm_df.loc[~is_na_row, self._feature_columns].values
+
+        y_proba_full = np.ones(len(psm_df))
+        y_proba = self._classifier.predict_proba(X)[:, 1]
+        y_proba_full[~is_na_row] = y_proba
+        return y_proba_full
+
+    def predict_qval(self, psm_df, fragments_df=None, dia_cycle=None):
```
mschwoer, line 104: docstring?

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+        if dia_cycle is not None and fragments_df is not None:
+            psm_df = _get_q_values(
+                psm_df,
+                score_column="decoy_proba",
+                decoy_column="decoy",
+                r_target_decoy=r_target_decoy,
+            )
+            fragment_competition = FragmentCompetition()
+            psm_df = fragment_competition(
+                psm_df[psm_df["qval"] < 0.10], fragments_df, dia_cycle
```
mschwoer, line 119: magic number?

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+        y_proba_full = np.ones(len(psm_df))
+        y_proba = self._classifier.predict_proba(X)[:, 1]
+        y_proba_full[~is_na_row] = y_proba
+        return y_proba_full
+
+    def predict_qval(self, psm_df, fragments_df=None, dia_cycle=None):
+        psm_df["decoy_proba"] = self.predict_classifier(psm_df)
+        # normalize to a 1:1 target decoy proportion
+        r_target_decoy = (psm_df["decoy"] == 0).sum() / (psm_df["decoy"] == 1).sum()
```
mschwoer, line 107: shouldn't this use `self._decoy_column`? (also at other parts of the module)

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+            psm_df, "decoy_proba", "decoy", r_target_decoy=r_target_decoy
+        )
+
+        # calulate PEP
+        psm_df["pep"] = _get_pep(
+            psm_df, score_column="decoy_proba", decoy_column="decoy"
+        )
+
+        _plot_fdr_curve(psm_df["qval"])
+        return psm_df
```
mschwoer, line 133: the dataframe is modified inplace, and returned .. I feel we should to only one thing (preferable the latter)

alphadia/fdrx/stats.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # numerical stability
+    EPSILON = 1e-6
+    pep = decoy_hist / (target_hist + decoy_hist + EPSILON)
+
+    return pep[np.digitize(score, score_bins) - 1]
+
+
+def _get_q_values(
+    _df: pd.DataFrame,
```
mschwoer, line 72: why the underscore? and mb give it a more speaking name?

alphadia/fdrx/stats.py
```diff
@@ -0,0 +1,166 @@
...
+        The q-values.
+    """
+    fdr_values_flipped = np.flip(fdr_values)
+    q_values_flipped = np.minimum.accumulate(fdr_values_flipped)
+    q_vals = np.flip(q_values_flipped)
+    return q_vals
+
+
+def _keep_best(
+    df: pd.DataFrame,
```
mschwoer, line 136: more speaking name? you can then use `df` instead of `temp_df` within the method

tests/unit_tests/test_fdrx_base.py
```diff
@@ -0,0 +1,45 @@
...
+        classifier=classifier_mock,
+        feature_columns=feature_columns,
+        competition_columns=["mod_seq_charge_hash"],
+    )
+
+    df = target_decoy_fdr.fit_predict_qval(mock_df)
+
+    # then
+    assert all([col in df.columns for col in ["decoy_proba", "qval", "pep"]])
+    assert np.all(df[["decoy_proba", "qval", "pep"]].values >= 0)
```
mschwoer, line 44: could we check for the actual values here or is it nondeterministic?

tests/unit_tests/test_fdrx_ploting.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
```
mschwoer, line 1: (nit) type in module name

mschwoer (approved):
LGTM

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+        np.ndarray
+            The decoy probabilities for the PSMs with same shape and order as the input dataframe.
+        """
+
+        is_na_row = psm_df[self._feature_columns].isna().any(axis=1)
+        X = psm_df.loc[~is_na_row, self._feature_columns].values
+
+        y_proba_full = np.ones(len(psm_df))
+        y_proba = self._classifier.predict_proba(X)[:, 1]
+        y_proba_full[~is_na_row] = y_proba
```
ammarcsj, line 101: maybe small comment why the default of setting to 1 for is appropriate for the na rows

alphadia/fdrx/stats.py
```diff
@@ -0,0 +1,166 @@
...
+    decoy_hist = np.convolve(decoy_hist, kernel_gaussian, mode="same")
+
+    # numerical stability
+    EPSILON = 1e-6
+    pep = decoy_hist / (target_hist + decoy_hist + EPSILON)
+
+    return pep[np.digitize(score, score_bins) - 1]
+
+
+def _get_q_values(
```
ammarcsj, line 71: would "add_q_values" be more descriptive?

alphadia/fdrx/stats.py
```diff
@@ -0,0 +1,166 @@
...
+    Returns
+    -------
+
+    pd.DataFrame
+        The dataframe containing the q-values in column qval.
+
+    """
+    EPSILON = 1e-6
+    _df = _df.sort_values([score_column, score_column], ascending=True)
+    target_values = 1 - _df[decoy_column].values
```
ammarcsj, line 106: can you add a quick reasoning why decoy values become targets if you do 1-? Or maybe improving naming could make it less confusing

alphadia/fdrx/base.py
```diff
@@ -0,0 +1,137 @@
...
+import pandas as pd
+import numpy as np
+from alphadia.fragcomp import FragmentCompetition
+from alphadia.fdrx.stats import _fdr_to_q_values, _get_pep, _get_q_values, _keep_best
+from alphadia.fdrx.plotting import (
+    _plot_fdr_curve,
+    _plot_roc_curve,
+    _plot_score_distribution,
+)
+
```
ammarcsj, line 14: What is your definition of a private function? One could argue that if you need to import these functions here, they should not be private anymore.

ammarcsj (approved):
Beautiful. Some small comments attached.

alphadia/fdrx/base.py
GeorgWa: The plots end up in the jupyter notebook, so this only has an effect during debugging.
Making sure thgat plot's end up on disk would be one of the open tasks.

alphadia/fdrx/stats.py
GeorgWa: Good point, add sounds more inplace

alphadia/fdrx/base.py
GeorgWa: Good point, this is really soft. The idea is that these functions should not be used outside the python module and there can be breaking changes any time.

PR #233, opened 2024-06-17
230 basic linting rule e

mschwoer:
- Add ruff linting to pre-commit
- add rule "E" to ruleset
- make code pass rule "E" (see individual commits)

### DIFF
--- a/alphadia/plexscoring.py
+++ b/alphadia/plexscoring.py
@@ -585,14 +585,14 @@ class Candidate:
         for i in range(_dense_precursors.shape[1]):
             dense_precursors[0, i, 0] = np.sum(_dense_precursors[0, i], axis=0)
             for k in range(_dense_precursors.shape[3]):
-                for l in range(_dense_precursors.shape[4]):
+                for ll in range(_dense_precursors.shape[4]):
                     sum = 0
                     count = 0
                     for j in range(_dense_precursors.shape[2]):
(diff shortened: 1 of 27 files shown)


alphadia/plexscoring.py
```diff
@@ -584,14 +585,14 @@ def process(
...
             dense_precursors[0, i, 0] = np.sum(_dense_precursors[0, i], axis=0)
             for k in range(_dense_precursors.shape[3]):
-                for l in range(_dense_precursors.shape[4]):
+                for ll in range(_dense_precursors.shape[4]):
                     sum = 0
                     count = 0
                     for j in range(_dense_precursors.shape[2]):
-                        sum += _dense_precursors[1, i, j, k, l]
-                        if _dense_precursors[1, i, j, k, l] > 0:
+                        sum += _dense_precursors[1, i, j, k, ll]
```
GeorgWa, line 592: Do you know what rule this is?

GeorgWa (approved):
LGTM, should not cause any issues

alphadia/plexscoring.py
mschwoer: https://www.flake8rules.com/rules/E741.html

I am really curious if each of these rules was set due to at least one big bug that caused a lot of trouble and them someone went like "meh, let's just forbid it"

PR #237, opened 2024-06-19
230 basic linting rule up

mschwoer:
- add rule "UP" to ruleset
- make code pass rule "UP" (see individual commits)


GeorgWa (approved):
LGTM, why is it not encouraged to open a file with `'r'`?

mschwoer:
> LGTM, why is it not encouraged to open a file with `'r'`?

it's not, but it's default behaviour of `open`

PR #241, opened 2024-06-19
230 basic linting rule b2

mschwoer:
- add rule "B" to ruleset
   - make code pass rule "B" (see individual commits)
- exceptions:
"B006" => next PR
"B905" => need to check which zips can be made strict

### DIFF
--- a/alphadia/numba/fragments.py
+++ b/alphadia/numba/fragments.py
@@ -159,7 +159,7 @@ def slice(inst, slices):
 
             precursor = np.arange(len(slices), dtype=np.uint32)
 
-            for i, (start_idx, stop_idx, step) in enumerate(slices):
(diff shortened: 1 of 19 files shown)


alphadia/numba/fragments.py
```diff
@@ -159,7 +159,7 @@ def impl(inst, slices):
 
             precursor = np.arange(len(slices), dtype=np.uint32)
 
-            for i, (start_idx, stop_idx, step) in enumerate(slices):
```
GeorgWa, line 162: What is the idea of this rule? that it's marked if it's unused?

GeorgWa (approved):
LGTM

alphadia/numba/fragments.py
mschwoer: yes, that's part of this ruleset

Issue #230, opened 2024-06-17
Basic linting


mschwoer:
WHO
Developers
WHAT
Have properly linted code
WHY
Best practises, less overhead when developing

AC
- Code is linted once using ruff (with sensible defaults), all issues are fixed (either by code changes or by applying `noqa`)
- pre-commit hook is in place that takes care about the linting (also in CI)
- add default pre-commit hooks (whitespaces, line-endings)
cf. https://github.com/MannLabs/alphabase/issues/154

PR #242, opened 2024-06-19
#230: ruff rule B006: auto fixes

mschwoer:
applying the "no mutable args" rule


GeorgWa (approved):
LGTM, may all bugs perish under this rule :D

PR #243, opened 2024-06-20
Add test split

mo-sameh:
Testing all three models on a test dataset only once after finetuning.


alphadia/transferlearning/train.py
```diff
@@ -432,7 +436,9 @@ def finetune_ms2(
 
         # Shuffle the psm_df and split it into train and test
         train_psm_df = psm_df.sample(frac=self.settings["train_ratio"]).copy()
-        test_psm_df = psm_df.drop(train_psm_df.index).copy()
+        rest_psm_df = psm_df.drop(train_psm_df.index).copy()
```
GeorgWa, line 439: (nit) typo in test

alphadia/transferlearning/train.py
```diff
@@ -32,7 +32,8 @@
     # --------- USer settings ------------
     "batch_size": 1000,
     "max_lr": 0.0005,
-    "train_ratio": 0.8,
+    "train_ratio": 0.7,
```
GeorgWa, line 35: I know its not required for calculation but let's add also test_ratio. What do you think about calling it `_fraction` instead of ratio? Is it really a rati

alphadia/transferlearning/train.py
```diff
@@ -432,7 +436,9 @@ def finetune_ms2(
 
         # Shuffle the psm_df and split it into train and test
         train_psm_df = psm_df.sample(frac=self.settings["train_ratio"]).copy()
-        test_psm_df = psm_df.drop(train_psm_df.index).copy()
+        rest_psm_df = psm_df.drop(train_psm_df.index).copy()
+        val_psm_df = rest_psm_df.sample(frac=self.settings["validation_ratio"]).copy()
```
GeorgWa, line 440: What do you think about adding a logger info with `test_ratio` and `n_test` for all three subsets?

alphadia/transferlearning/train.py
```diff
@@ -618,7 +650,7 @@ def finetune_rt(self, psm_df: pd.DataFrame) -> pd.DataFrame:
         # Create a callback handler
         callback_handler = CustomCallbackHandler(
             self._test_rt,
-            test_df=test_df,
+            test_df=val_df,
```
GeorgWa, line 653: Should it be val_df = val_df if we are strict? :D

alphadia/transferlearning/train.py
```diff
@@ -605,7 +634,10 @@ def finetune_rt(self, psm_df: pd.DataFrame) -> pd.DataFrame:
 
         # Shuffle the psm_df and split it into train and test
         train_df = psm_df.sample(frac=self.settings["train_ratio"])
-        test_df = psm_df.drop(train_df.index)
+        rest_df = psm_df.drop(train_df.index)
```
GeorgWa, line 637: Why is there rest_df and test_df ?

GeorgWa (approved):
LGTM

PR #246, opened 2024-06-20
#230: ruff rule B905: add strict keyword to zip

mschwoer:
As discussed: use "strict=True" for all* zips.

*except the one in the njit-decorated method

### DIFF
--- a/alphadia/features.py
+++ b/alphadia/features.py
@@ -22,7 +22,7 @@ def center_of_mass(
     scans, frames = np.nonzero(single_dense_representation > 0)
     if len(scans) == 0:
         return 0, 0
-    for scan, frame in zip(scans, frames):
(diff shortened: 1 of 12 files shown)


alphadia/features.py
GeorgWa, line 1: zip( strict=True) is not yet supported by numba https://github.com/numba/numba/issues/8943
I will mark the files

alphadia/peakgroup/search.py
GeorgWa, line 1: numba

GeorgWa (changes requested):

GeorgWa (approved):
LGTM

jalew188:
This should be applied to all AlphaX if necessary

GeorgWa (approved):

Issue #248, opened 2024-06-21
Update learning rate scheduler API


GeorgWa:
WHO
alphaDIA User

WHAT
Update transfer learning to new API

WHY
Warning messages, potential breaks in the future

Acceptance Criteria
No more warning messages :D

Additional information

```
0:12:48.068767 PROGRESS:  Ms2 model tested on validation dataset with the following metrics:
0:12:48.069153 PROGRESS:  l1_loss                       : 0.0411
0:12:48.069349 PROGRESS:  PCC-mean                      : 0.5771
0:12:48.069501 PROGRESS:  COS-mean                      : 0.6112
0:12:48.069706 PROGRESS:  SA-mean                       : 0.4187
0:12:48.069860 PROGRESS:  SPC-mean                      : 0.4531
0:12:48.071355 PROGRESS:  Fine-tuning MS2 model
[/Users/georgwallmann/miniconda3/envs/alpha/lib/python3.11/site-packages/torch/optim/lr_scheduler.py:28](https://file+.vscode-resource.vscode-cdn.net/Users/georgwallmann/miniconda3/envs/alpha/lib/python3.11/site-packages/torch/optim/lr_scheduler.py:28): UserWarning: The verbose parameter is deprecated. Please use get_last_lr() to access the learning rate.
  warnings.warn("The verbose parameter is deprecated. Please use get_last_lr() "
[/Users/georgwallmann/miniconda3/envs/alpha/lib/python3.11/site-packages/torch/optim/lr_scheduler.py:156](https://file+.vscode-resource.vscode-cdn.net/Users/georgwallmann/miniconda3/envs/alpha/lib/python3.11/site-packages/torch/optim/lr_scheduler.py:156): UserWarning: The epoch parameter in `scheduler.step()` was not necessary and is being deprecated where possible. Please use `scheduler.step()` to step the scheduler. During the deprecation, if epoch is different from None, the closed form is used instead of the new chainable form, where available. Please open an issue if you are unable to replicate your use case: https://github.com/pytorch/pytorch/issues/new/choose.
  warnings.warn(EPOCH_DEPRECATION_WARNING, UserWarning)
[/Users/georgwallmann/Documents/git/alphadia/alphadia/transferlearning/train.py:144](https://file+.vscode-resource.vscode-cdn.net/Users/georgwallmann/Documents/git/alphadia/alphadia/transferlearning/train.py:144): RuntimeWarning: invalid value encountered in scalar divide
[...]

PR #252, opened 2024-06-24
fix warnings for scheduler step and early stopping

mo-sameh:
Fixed 3 causes of warnings for the transfer learning module:
1) Step deprecated interface of passing epoch #248 
2) Verbose depreciated argument in the initialization of the the PyTorch ReduceLrOnPlateau. 
3) Early stopping first iteration invalid value encountered in scalar divide.


mschwoer (approved):
LGTM

PR #254, opened 2024-06-25
Include raw name in transfer lib

GeorgWa:
This PR makes sure the `raw_name` column is part of the transfer library


mschwoer (approved):

alphadia/outputaccumulator.py
```diff
@@ -134,6 +134,10 @@ def parse_output_folder(
         psm_df = psm_df[selected_precursor_columns]
         # validate.precursors_flat_from_output(psm_df)
 
+        # get foldername of the output folder
+        foldername = os.path.basename(folder)
+        psm_df["raw_name"] = foldername
```
jalew188, line 139: MB it is better to use a common get_raw_name() function

PR #255, opened 2024-06-25
FIX Overwrite old profile

GeorgWa:
So far, alphaDIA has loaded profile files, even if the version was mismatching.
This has led to errors when old profiles where missing config settings.

### DIFF
--- a/gui/src/main/modules/profile.js
+++ b/gui/src/main/modules/profile.js
@@ -1,13 +1,14 @@
-
 const fs = require("fs")
 const path = require("path")
 const { app, shell} = require("electron")
 const { dialog } = require('electron')
 
+const VERSION = "1.7.0"
+
 const Profile = class {
 
     config = {
-        "version": "1.7.0",
+        "version": VERSION,
         "conda": {
             "envName": "alpha",
             "path": ""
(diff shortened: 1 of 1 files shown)


gui/src/main/modules/profile.js
```diff
@@ -53,6 +52,20 @@ const Profile = class {
...
                 console.log(err)
             })
+            return
+        }
+
+        if (config.version !== VERSION) {
+            dialog.showMessageBox({
+                type: 'info',
+                title: 'Found old alphaDIA profile',
+                message: `Found old alphaDIA profile. Updating to version ${VERSION}`,
```
mschwoer, line 62:
```suggestion
                message: `Found old alphaDIA profile version ${config.version}. Updating to version ${VERSION}`,
```

gui/src/main/modules/profile.js
```diff
@@ -1,13 +1,14 @@
-
 const fs = require("fs")
 const path = require("path")
 const { app, shell} = require("electron")
 const { dialog } = require('electron')
 
+const VERSION = "1.7.0"
```
mschwoer, line 6: will this be bumped by bumpversion?

mschwoer (approved):

gui/src/main/modules/profile.js
GeorgWa: Yes

Issue #261, opened 2024-07-01
Docker Container Missing MANIFEST.in


dnusinow:
Thank you for alphadia and congratulations on the release!

**Describe the bug**
The alphadia docker container errors with the following error:

`ERROR: [Errno 2] No such file or directory: '/usr/local/lib/python3.9/site-packages/alphadia/constants/default.yaml'`. 

This was fixed by adding `COPY MANIFEST.in MANIFEST.in` to the Dockerfile around line 40.

**To Reproduce**
I followed the instructions [here](https://github.com/MannLabs/alphadia/blob/main/docs/installation.md#1-setting-up-docker) which worked perfectly with the patch suggested above.

**Expected behavior**
The alphadia run should go to completion.

**Logs**
I don't have the log anymore after patching the dockerfile.

**Version (please complete the following information):**
 - Installation Type Docker both from dockerhub and custom built from a git checkout
 - If no log is available, provide the following:
   - Platform information
     - system    Ubuntu linux
     - release   22.04.4 LTS (jammy)
     - version   [e.g. 10.15.7]
     - machine   amd64
     - processor Intel x86
     - cpu count 64
   - Python information:
     - alphadia version 1.7.0 release and [the current git head](https://github.com/MannLabs/alphadia/commit/3c3090c35623e1bced65337a3ba179046636d5dd).

dnusinow:
No problem, done!

PR #262, opened 2024-07-01
#261 install MANIFEST.in in docker container

dnusinow:
This is a PR for bug #261 . It installs the MANIFEST.in file in the docker container, which allows the pip install command to include all the files in the alphadia source directory, including the default configuration yaml, not just the .py files.


mschwoer (approved):
LGTM, thanks for the quick fix!

PR #265, opened 2024-07-05
Simplified pyinstaller script

GeorgWa:
Simplified pyinstaller script to honor hooks correctly (see https://github.com/pyinstaller/pyinstaller/issues/8614)

### DIFF
--- a/release/windows/build_backend.ps1
+++ b/release/windows/build_backend.ps1
@@ -1,9 +1,6 @@
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue ./build
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue ./dist
 
-pip install git+https://github.com/huggingface/huggingface_hub.git
-python -c 'from huggingface_hub import get_full_repo_name; print("success")'
-
 pip install build
 python -m build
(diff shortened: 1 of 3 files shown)


release/windows/build_backend.ps1
```diff
@@ -1,9 +1,6 @@
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue ./build
 Remove-Item -Recurse -Force -ErrorAction SilentlyContinue ./dist
 
-pip install git+https://github.com/huggingface/huggingface_hub.git
```
mschwoer, line 4: just wondering, we don't need the `huggingface_hub` dependenciy at all here?

mschwoer (approved):

release/windows/build_backend.ps1
jalew188: our DL models needs huggingface `transformers` and it needs `huggingface_hub`. We are going to remove the dependencies, see https://github.com/MannLabs/alphapeptdeep/issues/178

jalew188 (approved):

release/windows/build_backend.ps1
GeorgWa: The dependency is still Importen but just like all other dependencies. This Statement was included to get a Bugfix early which resolved some Import magic

PR #269, opened 2024-07-09
268 divisionbyzeroerror

mschwoer:
- fix division by zero error
- give some hint in case a bruker DDA file is opened
- fix issues with E2E cache cleanup

### DIFF
--- a/alphadia/data/bruker.py
+++ b/alphadia/data/bruker.py
@@ -81,7 +81,7 @@ class TimsTOFTranspose(alphatims.bruker.TimsTOF):
             )
             self.bruker_hdf_file_name = bruker_d_folder_name
         else:
-            raise NotImplementedError("WARNING: file extension not understood")
+            raise NotImplementedError("ERROR: file extension not understood")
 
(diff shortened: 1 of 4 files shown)


alphadia/data/bruker.py
```diff
@@ -65,13 +65,20 @@ def __init__(
...
-                    )
-                    raise ValueError(
-                        "Unexpected cycle shape. Will only retain first frame group"
+                        "Could not find cycle shape. Please check if this is a valid DIA data set."
                     )
+                    raise e
+                else:
+                    if cycle_shape != 1:
+                        msg = (
+                            "Unexpected cycle shape. Will only retain first frame group"
```
mschwoer, line 78: @GeorgWa The error message sounds like processing will continue, but apparently it does not .. shall we rephrase it?

GeorgWa: Yes, let's rephrase this to make it clear that we fail here.

GeorgWa (approved):
LGTM

Issue #260, opened 2024-07-01
Mono runtime error


rodvrees:
Hi all, 

I've been trying out AlphaDIA (pip installed, as per the directions, including pythonnet and mono) on both orbitrap and timsTOF diaPASEF data with public files from [PXD028735](https://www.ebi.ac.uk/pride/archive/projects/PXD028735).
The searches on orbitrap .raw files have been running smoothly, however while running AlphaDIA on the diaPASEF data I repeatedly encounter Mono runtime errors.

Here's my log:
[log.txt](https://github.com/user-attachments/files/16052161/log.txt)

I've put the thread count at 1 here, but I've attempted the search multiple times with different thread counts and GPU enabled/not enabled, and on different devices, and they all result in the same error (timepoints at which the error appears differ).

Some environment details:
- python 3.10.13
- pip 23.0.1
- mono 6.12.0.182
- conda 24.5.0

Here's the error:
===================================================
        Native Crash Reporting
=========================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=========================================

=========================================
        Native stacktrace:
=========================================
        0x7f5088933c13 - /home/robbe/miniconda3/envs/alphadia/lib/libmonosgen-2.0.so.1 : 
        0x7f5088933fc9 - /home/robbe/miniconda3/envs/alphadia/lib/libmonosgen-2.0.so.1 : 
        0x7f50888c60bb - /home/robbe/miniconda3/envs/alphadia/lib/libmonosgen-2.0.so.1 : 
        0x7f5088835a5b - /home/robbe/miniconda3/envs/alphadia/lib/libmonosgen-2.0.so.1 : 
        0x7f515d038050 - /lib/x86_64-linux-gnu/libc.so.6 : 
        0x7f506d1f8028 - Unknown
        0x7f506d1da8c2 - Unknown
        0x7f50694743fa - Unknown
        0x7f5069423164 - Unknown
        0x7f50557b155d - Unknown
        0x7f50557b1d17 - Unknown
        0x7f50e3a9103c - /home/robbe/miniconda3/envs/alphadia/lib/python3.12/site-packages/numba/_dispatcher.cpython-312-x86_64-linux-gnu.so : 
        0x7f50e3a91382 - /home/robbe/miniconda3/envs/alphadia/lib/python3.12/site-packages/numba/_dispatcher.cpython-312-x86_64-linux-gnu.so : 
[...]

PR #272, opened 2024-07-12
Fix out of bounds segfault

GeorgWa:
This should capture out-of bounds based sefaults like https://github.com/MannLabs/alphadia/issues/260

### DIFF
--- a/alphadia/features.py
+++ b/alphadia/features.py
@@ -352,8 +352,19 @@ def center_envelope(x):
                 ) * 0.5
 
 
-@nb.njit
-def center_envelope_1d(x):
+@nb.njit(inline="always")
+def _odd_center_envelope(x: np.ndarray):
+    """
+    Applies an interference correction envelope to a collection of odd-length 1D arrays.
+    Numba function which operates in place.
+
+    Parameters
+    ----------
+    x: np.ndarray
+        Array of shape (a, b) where a is the number of arrays and b is the length of each array.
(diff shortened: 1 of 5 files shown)


alphadia/plexscoring.py
```diff
@@ -1895,6 +1895,9 @@ def __call__(
         )
         thread_count = 1 if debug else thread_count
 
+        if debug:
```
mschwoer, line 1898: could you gather this and the other two `if debug`s in l. 1893 and 1896 in one block?

alphadia/features.py
```diff
@@ -362,16 +373,76 @@ def center_envelope_1d(x):
...
+    ----------
+    x: np.ndarray
+        Array of shape (a, b) where a is the number of arrays and b is the length of each array.
+        It is mandatory that dimension b is even.
+
+    """
+    center_index_right = x.shape[1] // 2
+    center_index_left = center_index_right - 1
+
+    for a0 in range(x.shape[0]):
```
mschwoer, line 403: why `a0`? just `a` would be less noisy

mschwoer (approved):
LGTM

alphadia/features.py
GeorgWa: that's a pattern I reuse a lot across numba function for iterating axes. I would keep it as long as I don't remove it completely from the codebase.

PR #273, opened 2024-07-13
Cleanup unused numba functions

GeorgWa:


mschwoer (approved):
wow, that was quite some unused code

PR #275, opened 2024-07-13
Bugfixes in GUI

GeorgWa:
This PR solves a couple of minor GUI bugs:
- Message boxes we're blocking the background process and not async
- Old profile.json settings were not overwritten correctly for new installations
- Selection of custom PeptDeep models was missing from the settings

### DIFF
--- a/gui/src/renderer/components/ParameterInput.js
+++ b/gui/src/renderer/components/ParameterInput.js
@@ -1,11 +1,44 @@
 import * as React from 'react'
 import styled from '@emotion/styled'
 
-import { Box, Checkbox, FormControl, MenuItem, Select, Stack, Tooltip, Typography, TextField } from '@mui/material'
+import { Box, Chip, Button, Checkbox, FormControl, MenuItem, Select, Stack, Tooltip, Typography, TextField } from '@mui/material'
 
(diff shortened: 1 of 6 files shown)


gui/src/renderer/components/ParameterInput.js
```diff
@@ -1,12 +1,45 @@
...
+
+    const handleSelect = () => {
+        window.electronAPI.getSingleFolder().then((folder) => {
+            onChange(folder);
+        }).catch((err) => {
+            console.log(err);
+        })
+    }
+
+    const folderName = parameter.replace(/^.*[\\\/]/, '')
```
mschwoer, line 20: why do we need this?

mschwoer (approved):
not a frontend person but looks clean to me

gui/src/renderer/components/ParameterInput.js
GeorgWa: For the sake of simplicity, every parameter input receives the value as its stored in the config. For file or folder inputs we want to display a small "chip" to indicate that a file has been selected (Grey bubbles #134). This regex selects only the file or folder name for display.

PR #278, opened 2024-07-14
Fail transparently for failed output accumulation

GeorgWa:
Screenshot 2024-07-14 at 15 47 21

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -27,6 +27,12 @@ from alphadia.transferlearning.train import FinetuneManager
 logger = logging.getLogger()
 
 
+class OutputGenerationError(Exception):
+    """Raised when an error occurs during output generation"""
+
+    pass
+
+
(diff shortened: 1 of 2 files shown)


alphadia/outputtransform.py
```diff
@@ -509,20 +515,17 @@ def build_precursor_table(
...
                     logger.warning(e)
-                    run_df = pd.DataFrame()
-
-            psm_df_list.append(run_df)
 
         if len(psm_df_list) == 0:
-            logger.error("No psm files found, can't continue")
-            raise FileNotFoundError("No psm files found, can't continue")
+            logger.error("No psm files accumulated, can't continue")
+            raise OutputGenerationError("No psm files accumulated, can't continue")
```
mschwoer, line 528: this is a quite generic .. could we make this more specific, e.g. "`NoPsmFound`"? the benefit would be that the we could have fine-granular error codes, which makes alphaDIA automation better to control

mschwoer (approved):

PR #279, opened 2024-07-16
make error handling automation-friendly

mschwoer:
- introduce custom error classes
- centralize and unify error handling
- write macine-readable error information to `events.jsonl`

### DIFF
--- a/alphadia/cli.py
+++ b/alphadia/cli.py
@@ -15,6 +15,7 @@ import yaml
 # alphadia imports
 import alphadia
 from alphadia import utils
+from alphadia.exceptions import CustomError
 from alphadia.workflow import reporting
 
 logger = logging.getLogger()
@@ -325,7 +326,7 @@ def run(*args, **kwargs):
     try:
         import matplotlib
 
-        # important to supress matplotlib output
+        # important to suppress matplotlib output
         matplotlib.use("Agg")
 
         from alphadia.planning import Plan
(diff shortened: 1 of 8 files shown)


alphadia/planning.py
```diff
@@ -340,36 +342,19 @@ def run(
...
-                )
-                logger.error(
-                    "      For large libraries, try to reduce the library size and reduce the calibration MS1 and MS2 tolerance"
-                )
-                logger.error(
-                    "   3. There was a fundamental issue with search parameters"
-                )
+            except CustomError as e:
+                _log_exception_event(e, raw_name, workflow)
                 continue
```
mschwoer, line 347: just confirming: in the case of a business error, we continue, on other errors we fail, correct?

but if for 100 files, 2 have business errors, then the remaning 98 will be quanted and success will be reported, right? how will the user know that 2 files had errors?

GeorgWa: Yes, exactly. The two files will be appear as all zero rows from the stat output.

alphadia/cli.py
```diff
@@ -341,8 +342,15 @@ def run(*args, **kwargs):
         plan.run()
 
     except Exception as e:
-        import traceback
-
-        logger.info(traceback.format_exc())
-        logger.error(e)
-        sys.exit(1)
+        if isinstance(e, CustomError):
+            message = f"An error occurred: {e.error_code}: {e.msg} {e.detail_msg}"
```
GeorgWa, line 346: Ideally the CLI module should have as little logic as possible so that the search behaves the same from the CLI or a jupyter notebook. I'm not sure if uncatched errors can actually reach this point or are catched before anyways.

GeorgWa (approved):
LGTM

alphadia/cli.py
mschwoer: That's a good point .. I moved all non-CLI relevant to `planning`

GeorgWa (approved):

PR #282, opened 2024-07-22
Selecting classifier version

odespard:
I have added an instance variable, FDRManager.current_version, which increments at each call of the FDRManager.fit_predict() method and is included in the hash assigned to the classifier when saved. I have also added a version parameter to the FDRManager.fit_predict() method which allows the selection of which version of the classifier to use, and another method, FDRManager.get_current_version(), to get the current version. 

Note that at present the new FDRManager changes the workflow, as the base classifier will be used repeatedly with this code (as opposed to the classifier being iteratively updated with each run as before). This could be changed by making the default argument for the version parameter `None` and including a statement along the lines of `if version is None then version = self.current_version` in the method, but I thought it worthwhile discuss this before adding it.


alphadia/workflow/manager.py
```diff
@@ -483,6 +483,17 @@ def fit_predict(self, update_dict):
 
 
 class FDRManager(BaseManager):
+    """
+    Contains, updates and applies classifiers for target-decoy competitio-based false discovery rate (FDR) estimation.
+    --------------------------------
```
GeorgWa, line 488: We ususally add the init docstring below the init `__init__`.
Can you also move the `----` below `Parameters` ?

alphadia/workflow/manager.py
```diff
@@ -512,6 +523,7 @@ def fit_predict(
         df_fragments: None | pd.DataFrame = None,
         dia_cycle: None | np.ndarray = None,
         decoy_channel: int = -1,
+        version: None | int = -1,
```
GeorgWa, line 526: I think we don't want to allow None

alphadia/workflow/manager.py
```diff
@@ -657,14 +677,27 @@ def load_classifier_store(self, path=None):
                         torch.load(os.path.join(path, file))
                     )
 
-    def get_classifier(self, available_columns):
+    def get_classifier(self, available_columns, version):
```
GeorgWa, line 680: please add type annotations for arguments and move the heading below `Parameters` :)
Should we also make `version` a kwarg with default -1 ?

tests/unit_tests/test_workflow.py
```diff
@@ -296,3 +297,51 @@ def test_workflow_base():
...
+    fdr_manager = manager.FDRManager(FDR_TEST_FEATURES, FDR_TEST_BASE_CLASSIFIER)
+
+    assert fdr_manager.is_loaded_from_file is False
+    assert fdr_manager.is_fitted is False
+
+    assert fdr_manager.feature_columns == FDR_TEST_FEATURES
+    assert fdr_manager.classifier_base == FDR_TEST_BASE_CLASSIFIER
+
+
+@pytest.mark.slow
```
GeorgWa, line 329: Not necessary. You should be able to replace `BinaryClassifierLegacyNewBatching` with any scikit learn classifier (or even a mock) to speed it up.

tests/unit_tests/test_workflow.py
```diff
@@ -296,3 +297,51 @@ def test_workflow_base():
...
+        decoy_strategy="precursor",
+        competetive=False,
+        df_fragments=None,
+        dia_cycle=None,
+        # neptune_run=self.neptune
+    )
+
+    assert len(fdr_manager.classifier_store) == 2
+    assert fdr_manager.current_version == 0
+    assert manager.column_hash(FDR_TEST_FEATURES) in fdr_manager.classifier_store
```
GeorgWa, line 347: Can you add a testcase which 1) calls fit_predict twice 2) tests retrival of different versions 3) loads and saves and makes sure that retrival still works after saving.

tests/unit_tests/test_workflow.py
odespard: The FDRManager.fit_predict() method requires the classifier to have the method .from_state_dict().

alphadia/workflow/manager.py
```diff
@@ -619,7 +638,14 @@ def fit_predict(
...
 
         self.is_fitted = True
-        self.classifier_store[column_hash(available_columns)] = classifier
+
+        classifier_hash = column_hash(available_columns)
+        if classifier_hash not in self.classifier_store:
+            self.classifier_store[classifier_hash] = []
+
+        self._current_version += 1
+        self.classifier_store[column_hash(available_columns)].append(classifier)
```
mschwoer, line 647: consider using `defaultdict`:
```python
from collections import defaultdict

# in constructor (whereever self.classifier_store is currently instantiated)
self.classifier_store = defaultdict(list)

# here
classifier_hash = column_hash(available_columns)
self.classifier_store[classifier_hash].append(classifier)


```

alphadia/workflow/manager.py
```diff
@@ -657,14 +683,33 @@ def load_classifier_store(self, path=None):
...
+        Returns
+        ----------
+        object
+            Classifier object
+
+        """
         classifier_hash = column_hash(available_columns)
         if classifier_hash in self.classifier_store:
-            classifier = self.classifier_store[classifier_hash]
+            classifier = deepcopy(self.classifier_store[classifier_hash][version])
```
mschwoer, line 704: so `version` is rather `version_index`?

alphadia/workflow/manager.py
```diff
@@ -657,14 +683,33 @@ def load_classifier_store(self, path=None):
...
         else:
             classifier = deepcopy(self.classifier_base)
         return classifier
```
mschwoer, line 707: (nit) the `deepcopy` could be moved here

tests/unit_tests/test_workflow.py
odespard: It turns out that save_classifier_store relies on the current 1d implementation of the classifier_store. Is it OK if I modify this function too?

tests/unit_tests/test_workflow.py
odespard: Is there an alternative way to speed this up? Or is it better to leave this as it is?

alphadia/workflow/manager.py
odespard: I suppose so. It isn't used with any other meaning though. Do you think it is a problem?

tests/unit_tests/test_workflow.py
odespard: I have now modified this method. It only saves the version passed to it as an argument (default -1).

GeorgWa (approved):

PR #283, opened 2024-07-24
MS2 Optimizer

odespard:
This is a draft that does not change the working of the current code but starts a file called optimizer.py in the alphadia/workflow folder which will be used as the basis for further changes to the way in which calibration optimization is performed. A base optimizer class is defined and an MS2 optimizer class is defined based on the base optimizer. A unit test is added for the MS2 optimizer.


alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+
+# alpha family imports
+# third party imports
+import pandas as pd
+
+# alphadia imports
+from alphadia.workflow import reporting
+
+
+def check_if_optimum_found(method):
```
GeorgWa, line 14: This is a bit dangerous as we are changing the interface dynamically.

Nevertheless I like the idea with the decorator. Can you change it so it raises a custom exception?
have a look at the exceptions introduced by magnus in this PR: https://github.com/MannLabs/alphadia/pull/279

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+            return True
+        return method(self, *args, **kwargs)
+
+    return wrapper
+
+
+class BaseOptimizer:
+    def __init__(
+        self,
+        initial_tolerance: float,
```
GeorgWa, line 33: I think the kernel optimizer might have two values. Maybe a dict is better here.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        reporter: None | reporting.Pipeline | reporting.Backend = None,
+    ):
+        """This class serves as a base class for organizing the search parameter optimization process, which defines the tolerances used for search.
+
+        Parameters
+        ----------
+        initial_tolerance: float
+            The initial tolerance to start the optimization
+        """
+        self.proposed_new_tolerance = initial_tolerance
```
GeorgWa, line 43: let's rename it from `tolerance` to `parameter`.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        self.tolerances = []
+        self.round = -1
```
GeorgWa, line 45: Could we also look at `len(self.tolerances)`? 

In general, you always wan't to have a piece of information only in one place as there is the risk that it get's out of sync.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+            The filtered dataframe of precursors obtained after FDR correction.
+
+        proposed_new_tolerance: float
+            The tolerance which will be used for the next round of optimization (if optimization continues).
+
+        """
+        self.ids.append(len(precursors_df))
+        self.proposed_new_tolerance = proposed_new_tolerance
+
+    def end_of_optimization_message(self):
```
GeorgWa, line 135: no need to separate this into a function now that we remove it from the decorator.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class RTOptimizer(BaseOptimizer):
+    pass
+
+
+class MS2Optimizer(BaseOptimizer):
+    def __init__(self, initial_tolerance: float):
+        super().__init__(initial_tolerance)
+        self.estimator = "mz"
```
GeorgWa, line 70: is this used?

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        self.df = "fragment"
+        self.ids = []
```
GeorgWa, line 72: `ids` can mean a lot of things. Maybe `precursor_ids`.
I think we will also switch to a df here at some point so we can keep track of multiple parameters.

tests/unit_tests/test_workflow.py
```diff
@@ -369,3 +370,35 @@ def test_fdr_manager_fit_predict():
...
+    test_dict = defaultdict(list)
+    test_dict["var"] = list(range(100))
+
+    ms2_optimizer = optimizer.MS2Optimizer(100)
+
+    assert ms2_optimizer.optimal_tolerance is None
+    assert ms2_optimizer.round == -1
+    assert ms2_optimizer.check_stopping_conditions() is False
+
+    ms2_optimizer.initiate()
```
GeorgWa, line 385: I'm not quite sure why there is a `initiate` function. The naming is a bit ambigous with the object constructor `__init__`.

Would the following work:

```
if not optimizer.has_converged():
   optimizer.step() # or .update()
```

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    def update(self):
+        raise NotImplementedError(
+            f"update() not implemented for {self.__class__.__name__}"
+        )
+
+    def initiate(self):
+        raise NotImplementedError(f"initiate() not implemented for {self.__class__}")
+
```
GeorgWa, line 61: let's add a `has_converged` function here.
You can just return `self.optimal_tolerance is not None`.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+class MS2Optimizer(BaseOptimizer):
+    def __init__(self, initial_tolerance: float):
+        super().__init__(initial_tolerance)
+        self.estimator = "mz"
+        self.df = "fragment"
+        self.ids = []
+
+    @check_if_optimum_found
+    def check_stopping_conditions(self):
```
GeorgWa, line 75: this function is good but i think it should be private and called automatically after each call to `.update()`

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        return stop_optimizaton
+
+    @check_if_optimum_found
+    def initiate(self):
+        """If the optimization continues (which will use the proposed new tolerance), the round variable should be incremented and the proposed new tolerance should be added to the list of tolerances."""
+        self.tolerances.append(self.proposed_new_tolerance)
+        self.round += 1
+
+    @check_if_optimum_found
+    def update(self, precursors_df: pd.DataFrame, proposed_new_tolerance: float):
```
GeorgWa, line 118: I think this is the place where we want to calcualte the `proposed_new tolerance`.

GeorgWa (commented):
Very good OOP code! :)

Let's sharpen the object a bit. I think some names are not as clear as they can be and it's not immediately clear how the public functions `initiate, update, check_stopping_conditions` should be used.

I would propose to rename/resutructure them to:
```
if not optimizer.has_converged():
   optimizer.step() # or .update()
```

As a general note, I think we will give the optimizer access to the peptide centric workflow optimization manager so it can set values after each round and once it has converged and backtracks.

alphadia/workflow/optimizer.py
odespard: Yes, I will make it return an exception instead. In this case I will remove the decorator from `check_stopping_conditions` (which I will rename `has_converged` as suggested) so that this still returns True if called a subsequent time.

alphadia/workflow/optimizer.py
odespard: Do you mean replace self.round with len(self.tolerances)? That seems to make sense to me.

alphadia/workflow/optimizer.py
odespard: Indeed.

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+    """If an optimum has been found, the optimizer should cease to execute its methods to check, initiate and update.
+    This function can be used as a decorator and placed before the methods in an optimizer to prevent execution of the methods if an optimum has been found.
+    If the optimum has been found, the decorator will print the end of optimization message and, for consistency with stop_or_continue methods, return True.
+    """
+
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
+        if self.optimal_tolerance is not None:
+            self.end_of_optimization_message()
+            return True
```
mschwoer, line 24: I feel this is a bit of a dangerous case of how to use decorators because you don't know if the method you are wrapping is actually returning a boolean or something else ..
 .. still, I second of course your striving for less code duplication :-)

Think about this alternative:

add this stopping logic into `BaseOptimizer`:
```python
def update(self):
        if self.optimal_tolerance is not None:
            self.end_of_optimization_message()
            return
         self._update()

def _update(self):
        raise NotImplementedError(
            f"_update() not implemented for {self.__class__.__name__}"
        )

# subclasses implement `_update`
```
This way, you have the stopping logic centralized (although with a bit more code dup).

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        return method(self, *args, **kwargs)
+
+    return wrapper
+
+
+class BaseOptimizer:
```
mschwoer, line 30: inherit from `ABC` ?

alphadia/workflow/optimizer.py
```diff
@@ -0,0 +1,147 @@
...
+        raise NotImplementedError(
+            f"update() not implemented for {self.__class__.__name__}"
+        )
+
+    def initiate(self):
+        raise NotImplementedError(f"initiate() not implemented for {self.__class__}")
+
+
+class RTOptimizer(BaseOptimizer):
+    pass
```
mschwoer, line 64: this will be implemented later, I guess? could you this as a comment?

alphadia/workflow/optimizer.py
odespard: I see. I will implement this alongside a dictionary for initial_tolerance.

alphadia/workflow/optimizer.py
mschwoer: after reading this comment https://github.com/MannLabs/alphadia/pull/283/files#r1691005405 :
 could also be something like
```python
def update(self) -> bool:
# return value indicates wheter to stop
         self._update()
         return check_stopping_conditions()

def _update(self):
...
```

alphadia/workflow/optimizer.py
odespard: OK. I presume this should be applied throughout.

alphadia/workflow/optimizer.py
odespard: OK

alphadia/workflow/optimizer.py
mschwoer: some for the others

tests/unit_tests/test_workflow.py
odespard: There is an initiate function (which I will rename) so that calling the check function multiple times doesn't change the tolerance value. This issue arose during debugging and almost certainly won't be important for the actual use of the object, so I suppose there is no real need to separate the initiation from the check.

mschwoer: Just realizing another reason for writing and actually also reviewing unit tests: you can judge much better how the new interfaces work in practise than by just looking at the code

alphadia/workflow/optimizer.py
odespard: I am happy to do that. I didn't see ABC being used elsewhere in the code so I was trying to avoid it.

alphadia/workflow/optimizer.py
odespard: OK.

alphadia/workflow/optimizer.py
odespard: With respect to the decorator, what do you think of Georg's suggestion(https://github.com/MannLabs/alphadia/pull/283#discussion_r1690983718)? Albetít I think that if I combine the check_stopping_conditions and initiate methods as suggested elsewhere then the decorator is probably superfluous. 

I agree setting the return value is a bad idea.

alphadia/workflow/optimizer.py
odespard: Could I just set has_converged as an attribute at the same time as setting the optimal tolerance?

odespard: Or is it safer to have a method that checks optimal_tolerance explicitly?

alphadia/workflow/optimizer.py
mschwoer: I think the decorator will cause more trouble than good in the long run ;-)

alphadia/workflow/optimizer.py
odespard: I agree this would be preferable but at the moment the means of calculating a new tolerance requires the calibration_manager and I wasn't sure we wanted to create a dependency of the optimizer object on the calibration_manager. Do you think I should do this?

alphadia/workflow/optimizer.py
odespard: Not within the object, but I can't see how to avoid using it when setting the new parameter with calibration_manager (see https://github.com/MannLabs/alphadia/pull/283#discussion_r1691069702).

alphadia/workflow/optimizer.py
odespard: OK. This would be good as then it can still be called separately for debugging purposes, I suppose.

tests/unit_tests/test_workflow.py
odespard: Taking this approach will probably resolve the problem (https://github.com/MannLabs/alphadia/pull/283#discussion_r1691073315).

alphadia/workflow/optimizer.py
GeorgWa: Yes, that's fine. This class doesn't need to be general.

alphadia/workflow/optimizer.py
GeorgWa: A different strategy would be to remote this from the base class at all and make it an ABC.

odespard: I decided to remove it from the base class.

GeorgWa (approved):
LGTM

PR #290, opened 2024-07-29
raise on missing checkpoint path

GeorgWa:
Raise error on missing checkpoint path

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -41,7 +41,9 @@ library_prediction:
   # maximum charge state for predicted fragments
   max_fragment_charge: 2
   instrument: Lumos
-  checkpoint_folder_path: null
+
+  # set path for custom peptdeep model. If set to null, the default model will be used
+  peptdeep_model_path: null
 
 search:
   channel_filter: '0'
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -41,7 +41,7 @@ library_prediction:
   # maximum charge state for predicted fragments
   max_fragment_charge: 2
   instrument: Lumos
-  checkpoint_folder_path: None
+  checkpoint_folder_path: null
 
 search:
   channel_filter: '0'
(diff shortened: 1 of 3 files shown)


alphadia/constants/default.yaml
```diff
@@ -41,7 +41,7 @@ library_prediction:
   # maximum charge state for predicted fragments
   max_fragment_charge: 2
   instrument: Lumos
-  checkpoint_folder_path: None
+  checkpoint_folder_path: null
```
mschwoer, line 44: could you add a short comment here: what behaviour does 'null' mean, why would I set this?

also: really `null`?

alphadia/libtransform.py
```diff
@@ -314,6 +314,11 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
 
         model_mgr = ModelManager(device=device)
         if self.checkpoint_folder_path is not None:
+            if not os.path.exists(self.checkpoint_folder_path):
```
mschwoer, line 317: is `checkpoint` really a good name? I confused it with the `.progress` concept ..

alphadia/constants/default.yaml
GeorgWa: It's stupid, I agree. `null` in a yaml translates into `None` in python. `None` gets parsed as `'None'` 😭

alphadia/libtransform.py
GeorgWa: This is how it's called within peptdeep at the moment. We could name it `peptdeep_model_path`, what do you think?

mschwoer: sounds good

alphadia/constants/default.yaml
mschwoer: again what learned ..
could we change this to 
`checkpoint_folder_path: # empty: use default ~/peptdeep`

GeorgWa: Yep, thats the current behavior with `null`.
We couuuld, but in yaml `key: ` gets parsed as `key: null`.  😄 so maybe the verbose version is better?

mschwoer (approved):

PR #291, opened 2024-07-29
Add unknown mods

GeorgWa:
Allow users to set unknown alphabase mods by using the config.
Required minor refactoring of the config class to allow addition of keys not in the original config.


JuliaS92:
Just raising this question because I had an issue with doing the same thing in peptdeep today and trying to learn: How does this work with the multiprocessing on Mac? I had the issue, that I needed to modify the constants table on disk rather than temporarily in a jupyter notebook for library prediction and multiprocessing to recognize the change. Is this specifically a problem for Mac + jupyter or does it potentially affect other execution forms on Mac as well?

GeorgWa:
Hi @JuliaS92 . I think this a known issue with Jupyter and multiprocessing. It should work fine if it's in a python module.

PR #292, opened 2024-07-30
Multiplex spectral library

GeorgWa:
Allow automated multiplexing of spectral library from the config.yaml using a config:
```
# perform non-isobaric multiplexing of any input library
library_multiplexing:
  # if true, the library is multiplexed
  enabled: False

  # define channels by their name and how modifications should be translated from the input library to the multiplexed library
  multiplex_mapping:
    0:
      mTRAQ@K: mTRAQ@K
      mTRAQ@Any_N-term: mTRAQ@Any_N-term

    4:
      mTRAQ@K: mTRAQ:13C(3)15N(1)@K
      mTRAQ@Any_N-term: mTRAQ:13C(3)15N(1)@Any_N-term

    8:
      mTRAQ@K: mTRAQ:13C(6)15N(2)@K
      mTRAQ@Any_N-term: mTRAQ:13C(6)15N(2)@Any_N-term

    12:
      mTRAQ@K: mTRAQ:d12@K
      mTRAQ@Any_N-term: mTRAQ:d12@Any_N-term
```

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -163,23 +163,32 @@ library_multiplexing:
   # if true, the library is multiplexed
   enabled: False
 
+  # if the input library already contains multiplexed channels, the input channel has to be specified.
+  input_channel: 0
+
   # define channels by their name and how modifications should be translated from the input library to the multiplexed library
-  multiplex_mapping:
-    0:
-      mTRAQ@K: mTRAQ@K
-      mTRAQ@Any_N-term: mTRAQ@Any_N-term
+  # channels can be either a number or a string
+  # for every channel, the library gets copied and the modifications are translated according to the mapping
+  # the following example shows how to multiplex mTRAQ to three sample channels and a decoy channel
+  multiplex_mapping: {}
 
-    4:
-      mTRAQ@K: mTRAQ:13C(3)15N(1)@K
-      mTRAQ@Any_N-term: mTRAQ:13C(3)15N(1)@Any_N-term
+    #0:
+    #  mTRAQ@K: mTRAQ@K
+    #  mTRAQ@Any_N-term: mTRAQ@Any_N-term
 
-    8:
-      mTRAQ@K: mTRAQ:13C(6)15N(2)@K
-      mTRAQ@Any_N-term: mTRAQ:13C(6)15N(2)@Any_N-term
+    #4:
+    #  mTRAQ@K: mTRAQ:13C(3)15N(1)@K
+    #  mTRAQ@Any_N-term: mTRAQ:13C(3)15N(1)@Any_N-term
 
-    12:
-      mTRAQ@K: mTRAQ:d12@K
-      mTRAQ@Any_N-term: mTRAQ:d12@Any_N-term
+    #8:
+    #  mTRAQ@K: mTRAQ:13C(6)15N(2)@K
+    #  mTRAQ@Any_N-term: mTRAQ:13C(6)15N(2)@Any_N-term
+
+    #12:
+    #  mTRAQ@K: mTRAQ:d12@K
+    #  mTRAQ@Any_N-term: mTRAQ:d12@Any_N-term
+
+
 
 multiplexing:
   enabled: False
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -158,6 +158,29 @@ scoring_config:
   precursor_mz_tolerance: 10
   fragment_mz_tolerance: 15
 
+# perform non-isobaric multiplexing of any input library
+library_multiplexing:
+  # if true, the library is multiplexed
+  enabled: False
+
+  # define channels by their name and how modifications should be translated from the input library to the multiplexed library
+  multiplex_mapping:
+    0:
+      mTRAQ@K: mTRAQ@K
+      mTRAQ@Any_N-term: mTRAQ@Any_N-term
+
+    4:
+      mTRAQ@K: mTRAQ:13C(3)15N(1)@K
+      mTRAQ@Any_N-term: mTRAQ:13C(3)15N(1)@Any_N-term
+
+    8:
+      mTRAQ@K: mTRAQ:13C(6)15N(2)@K
+      mTRAQ@Any_N-term: mTRAQ:13C(6)15N(2)@Any_N-term
+
+    12:
+      mTRAQ@K: mTRAQ:d12@K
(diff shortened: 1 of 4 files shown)


alphadia/constants/default.yaml
ammarcsj, line 1: dimethyl multiplexing is not an option?

alphadia/constants/default.yaml
```diff
@@ -158,6 +158,29 @@ scoring_config:
...
   fragment_mz_tolerance: 15
 
+# perform non-isobaric multiplexing of any input library
+library_multiplexing:
+  # if true, the library is multiplexed
+  enabled: False
+
+  # define channels by their name and how modifications should be translated from the input library to the multiplexed library
+  multiplex_mapping:
+    0:
```
ammarcsj, line 168: (nit) these ints are basically the channel names, i.e. could this in theory also be an arbitrary string? Maybe a small comment on this, as I was initially confused by the ints

alphadia/libtransform.py
```diff
@@ -602,6 +604,58 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
+        return valid
+
+    def forward(self, input: SpecLibBase) -> SpecLibBase:
+        """Apply the MultiplexLibrary step to the input object."""
+
+        if "channel" in input.precursor_df.columns:
+            channel_unique = input.precursor_df["channel"].unique()
+            logger.info(f"Library already multiplexed for channel {channel_unique}")
+            input.precursor_df = input.precursor_df[input.precursor_df["channel"] == 0]
+
```
ammarcsj, line 636: The implicit assumption is that unmodified is always encoded as 0 (and not as "ref" of whatever)? Maybe then specify this also in the yaml, which relates to the comment above

ammarcsj: Is the case that a library e.g. has only channel 2 likely, and would the code behave as expected in this case?

ammarcsj (approved):
LGTM

alphadia/libtransform.py
```diff
@@ -602,6 +604,58 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
+                        valid = False
+
+        return valid
+
+    def forward(self, input: SpecLibBase) -> SpecLibBase:
+        """Apply the MultiplexLibrary step to the input object."""
+
+        if "channel" in input.precursor_df.columns:
+            channel_unique = input.precursor_df["channel"].unique()
+            logger.info(f"Library already multiplexed for channel {channel_unique}")
```
odespard, line 634: My understanding is that any channels (apart from 0) for which the library is already multiplexed are lost in the subsequent lines of code. If so, should the log reflect this?

odespard (approved):

alphadia/libtransform.py
```diff
@@ -602,6 +604,58 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
+        valid &= isinstance(input, SpecLibBase)
+
+        # check if all modifications are valid
+        for _, channel_multiplex_mapping in self._multiplex_mapping.items():
+            for key, value in channel_multiplex_mapping.items():
+                for mod in [key, value]:
+                    print(mod)
+                    if mod not in MOD_DF.index:
+                        logger.error(f"Modification {mod} not found in input library")
+                        valid = False
```
boopthesnoot, line 625: we're not exiting the loops here to show all `mod`s that are not in the library as not to play "error ping pong", as Magnus says? otherwise we could save some iterations here

alphadia/libtransform.py
```diff
@@ -602,6 +604,58 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
+                channel_lib._precursor_df["channel"] = channel
+
+            channel_lib.calc_fragment_mz_df()
+            channel_lib_list.append(channel_lib)
+
+        def apply_func(x, y):
+            x.append(y)
+            return x
+
+        speclib = reduce(lambda x, y: apply_func(x, y), channel_lib_list)
```
boopthesnoot, line 654: is this okay performance-wise, or do we need to create a concat method for speclibbase?

boopthesnoot (approved):

tests/unit_tests/test_libtransform.py
```diff
@@ -79,3 +81,53 @@ def test_library_transform():
...
+        assert (
+            result_lib.precursor_df[
+                result_lib.precursor_df["channel"] == channel
+            ].shape[0]
+            == repeat
+        )
+
+    for modification in ["mTRAQ@K", "mTRAQ:13C(3)15N(1)@K", "mTRAQ:13C(6)15N(2)@K"]:
+        assert (
+            result_lib.precursor_df[
```
vbrennsteiner, line 129: (nit) I think one could iterate over test_multiplex_mapping.items() and then slice the precursor_df, which would avoid repeating the contents of test_multiplex_mapping. If at some point the test is parametrized with flexible mapping inputs, these loops will be in the way.

vbrennsteiner (approved):

alphadia/libtransform.py
mschwoer: cf. https://github.com/MannLabs/alphabase/pull/199#discussion_r1680593095

alphadia/constants/default.yaml
GeorgWa: This is just an example, added a comment.

alphadia/libtransform.py
GeorgWa: I've added a parameter to select the input channel explicitly and added relveant warnings and checks

alphadia/libtransform.py
```diff
@@ -602,6 +604,58 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
...
+    def validate(self, input: str) -> bool:
+        """Validate the input object. It is expected that the input is a path to a file which exists."""
+        valid = True
+        valid &= isinstance(input, SpecLibBase)
+
+        # check if all modifications are valid
+        for _, channel_multiplex_mapping in self._multiplex_mapping.items():
+            for key, value in channel_multiplex_mapping.items():
+                for mod in [key, value]:
+                    print(mod)
```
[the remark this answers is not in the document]
GeorgWa: debugging ;)

alphadia/libtransform.py
GeorgWa: Yes, exactly. The `validate` method should collect and print errors and fail afterwards

alphadia/libtransform.py
GeorgWa: I think that's fine :)

PR #293, opened 2024-07-30
Add automatic calibration

odespard:


alphadia/constants/default.yaml
```diff
@@ -95,17 +95,20 @@ search_advanced:
 
 calibration:
 
-  # minimum number of times (epochs) the updated calibration target has to been passed
-  min_epochs: 3
+  # minimum number of times the classifier is trained.
+  min_training_iterations: 3
```
GeorgWa, line 99: I think the naming and comment are not very precise. Can we rename it to something similar to min_steps_optimization_lock. And emphasize that this is for acquiring the optimization lock/subset. I think we also need a concise name for this process and use it consistently.

alphadia/constants/default.yaml
```diff
@@ -95,17 +95,20 @@ search_advanced:
...
 
   # Number of precursors searched and scored per batch
   batch_size: 8000
 
-  # recalibration target for the first epoch. For subsequent epochs, the target will increase by this amount.
-  recalibration_target: 200
+  # minimum number of precursors to be found before search parameter optimization begins
+  min_precursors_for_optimization: 200
```
GeorgWa, line 105: Same to above, let's call it optimization_lock_target or something

alphadia/workflow/peptidecentric.py
```diff
@@ -15,9 +15,8 @@
 
 # alphadia imports
 from alphadia import fragcomp, plexscoring, utils
-from alphadia.exceptions import NoRecalibrationTargetError
 from alphadia.peakgroup import search
-from alphadia.workflow import base, manager
+from alphadia.workflow import base, manager, searchoptimization
```
GeorgWa, line 19: let's call it just `optimization` or similar, I think the `search` doesn't ad much.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,119 +314,85 @@ def get_batch_plan(self):
 
         return plan
 
-    def start_of_calibration(self):
-        self.batch_plan = self.get_batch_plan()
+    def extract_optimization_data(self, target, min_iterations):
```
GeorgWa, line 317: Good idea to put this into in a separate method.
Can we align the name more so it becomes clear what process this is?

Can youy check if we need to pass `target` and `min_iterations` as arguments?
Otherwise we could read them directly from the log.
(Nit) lets give them the same name as the parameters in the log.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,119 +314,85 @@ def get_batch_plan(self):
...
+                verbosity="progress",
+            )
 
-        self.reporter.log_string(
-            f"=== Epoch {self.com.current_epoch}, step {current_step}, extracting elution groups {start_index} to {stop_index} ===",
-            verbosity="progress",
-        )
+            eg_idxes = self.elution_group_order[start_index:stop_index]
+            batch_df = self.spectral_library._precursor_df[
+                self.spectral_library._precursor_df["elution_group_idx"].isin(eg_idxes)
```
GeorgWa, line 348: This is just a comment/thought. I did this as a "cheap" way to get random elution groups from the library. In hindsight this operation is quite expensive for a large library. We can think of strategies how to do this different in the future.

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
             )
             return
 
-        self.start_of_calibration()
-        for current_epoch in range(self.config["calibration"]["max_epochs"]):
-            if self.check_epoch_conditions():
-                pass
+        if self.config["search"]["target_ms2_tolerance"] > 0:
```
GeorgWa, line 408: Very clean and understandable code!
Can you move this code up to 486 into a separate function which returns the optimizer list?

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
...
+        )
+
+        self.com.fit({"classifier_version": self.fdr_manager.current_version})
+
+        precursor_df_filtered, fragments_df_filtered = self.filter_dfs(
+            precursor_df, fragments_df
+        )
 
-            self.start_of_epoch(current_epoch)
+        self.recalibration(precursor_df_filtered, fragments_df_filtered)
```
GeorgWa, line 505: Just to understand it correctly, this is the last calibration as part of getting the optimization lock?

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
...
+            "Required number of precursors found and required number of training iterations performed. Starting search parameter optimization.",
+            verbosity="progress",
+        )
 
-                eg_idxes = self.elution_group_order[start_index:stop_index]
+        self.reporter.log_string(
+            "=== Performing initial optimization on extracted data. ===",
+            verbosity="info",
+        )
+        # Perform an initial optimization step based on the extracted data to update the initial search parameters.
```
GeorgWa, line 516: I don't understand that part.

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
...
+        for optimizers in order_of_optimization:
+            for current_step in range(self.config["calibration"]["max_steps"]):
+                if np.all([optimizer.has_converged for optimizer in optimizers]):
+                    self.reporter.log_string(
+                        f"Optimization finished for {', '.join([optimizer.parameter_name for optimizer in optimizers])}.",
+                        verbosity="progress",
+                    )
+
+                    for optimizer in optimizers:
+                        if isinstance(optimizer, searchoptimization.AutomaticOptimizer):
```
GeorgWa, line 531: I think we don't need to check the is instance. Let's just implement an empty `.plot` method for Target optimizers

alphadia/workflow/peptidecentric.py
```diff
@@ -546,9 +623,24 @@ def recalibration(self, precursor_df, fragments_df):
             ),
             max_fragments,
         )
-        fragments_df_filtered = fragments_df_filtered.iloc[:stop_rank]
+        fragments_df_filtered = fragments_df_filtered.iloc[
+            :stop_rank
+        ]  # QUESTION: Should this raise an exception if the length of fragments_df_full is less than min_fragments?
```
GeorgWa, line 628: No, that's fine

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
...
+                        f"Optimization finished for {', '.join([optimizer.parameter_name for optimizer in optimizers])}.",
+                        verbosity="progress",
+                    )
+
+                    for optimizer in optimizers:
+                        if isinstance(optimizer, searchoptimization.AutomaticOptimizer):
+                            optimizer.plot()
+
+                    break
                 batch_df = self.spectral_library._precursor_df[
```
GeorgWa, line 535: do we need to regenerate this every iteration? isn't this frozen for the whole optimization?

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +405,205 @@ def calibration(self):
...
                     self.spectral_library._precursor_df["elution_group_idx"].isin(
-                        eg_idxes
+                        self.eg_idxes_for_calibration
```
GeorgWa, line 537: Can you check if this needs to be an instance variable?

alphadia/workflow/searchoptimization.py
```diff
@@ -58,8 +49,140 @@ def step(self, precursors_df: pd.DataFrame, fragments_df: pd.DataFrame):
         """
         pass
 
+
+class AutomaticOptimizer(BaseOptimizer):
+    def __init__(
+        self,
+        initial_parameter: float,
+        workflow,
```
GeorgWa, line 57: maybe add the Args from BaseOptimizer explicitly so we know what they are and they become part of type hints, docstrings.

alphadia/workflow/searchoptimization.py
```diff
@@ -58,8 +49,140 @@ def step(self, precursors_df: pd.DataFrame, fragments_df: pd.DataFrame):
...
+            self.reporter.log_string(
+                f"✅ {self.parameter_name:<15}: optimization complete. Optimal parameter {self.workflow.com.__dict__[self.parameter_name]} found after {len(self.history_df)} searches.",
+                verbosity="progress",
+            )
+            return
+
+        new_row = pd.DataFrame(
+            [
+                {
+                    "parameter": float(
```
GeorgWa, line 91: Are the type conversions needed? this is rather uncommon for python and is usually a sign that something is off.

GeorgWa (commented):
Looks very good!
Can you take a look at all docstrings and type annotations and make sure they are up to date.

I think we also need to improve the logging a bit, as it wasn't great before. But we can figure this out.

alphadia/workflow/peptidecentric.py
odespard: OK.

alphadia/workflow/searchoptimization.py
odespard: No, I have removed them.

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +485,126 @@ def calibration(self):
             )
             return
 
-        self.start_of_calibration()
-        for current_epoch in range(self.config["calibration"]["max_epochs"]):
-            if self.check_epoch_conditions():
-                pass
-            else:
-                break
+        order_of_optimization = self.get_optimizers()
```
GeorgWa, line 488: should we now rename this to `optimizer_list` or similar?

alphadia/workflow/peptidecentric.py
GeorgWa: okay... let's keep this for now but i guess there is some future optimization potential

alphadia/workflow/peptidecentric.py
GeorgWa: I think I get it now :D

alphadia/workflow/peptidecentric.py
odespard: Perhaps `ordered_optimizers`?

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+
+        fragments_df: pd.DataFrame
+            The fragment dataframe for the search
+
+
+        """
+        pass
+
+
+class TargetedOptimizer(BaseOptimizer):
```
mschwoer, line 256: `FixedToleranceOptimizer`? and rename `parameter` -> `tolerance`? (mb I am completely off)

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+            The current step in the optimization process. By default it is set to -1, which prevents the optimizer from converging unless min_steps has been set to 0.
+
+        Returns
+        -------
+        bool
+            True if the convergence conditions are met, False otherwise.
+
+        """
+
+        return (
```
mschwoer, line 397: (nit)

```python

        min_steps_reached = current_step >= self.workflow.config["calibration"]["min_steps"] - 1
        return (
            min_steps_reached and
            len(self.history_df) > 2
            and self.history_df[self.feature_name].iloc[-1]
            < 1.1 * self.history_df[self.feature_name].iloc[-2]
            and self.history_df[self.feature_name].iloc[-1]
            < 1.1 * self.history_df[self.feature_name].iloc[-3]
            
        )
```
1. easier on the eye (imho)
2. if not `min_steps_reached`, the rest of the calculations is skipped (another fistful of CPU cycles saved ;-))

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+            len(self.history_df) > 2
+            and self.history_df[self.feature_name].iloc[-1]
+            < 1.1 * self.history_df[self.feature_name].iloc[-2]
+            and self.history_df[self.feature_name].iloc[-1]
+            < 1.1 * self.history_df[self.feature_name].iloc[-3]
+            and current_step >= self.workflow.config["calibration"]["min_steps"] - 1
+        )
```
mschwoer, line 471: try to move this logic up in the class hierarchy (duplicated 5 times)

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+
+        Returns
+        -------
+        float
+            The proposed new value for the search parameter.
+
+        """
+        return 1.1 * self.workflow.calibration_manager.get_estimator(
+            self.estimator_group_name, self.estimator_name
+        ).ci(df, 0.99)
```
mschwoer, line 488: also here: code duplication

mschwoer (commented):
that was a lot to digest ;-)

alphadia/workflow/peptidecentric.py
```diff
@@ -319,121 +282,198 @@ def get_batch_plan(self):
...
+
+        eg_idxes_for_calibration = self.elution_group_order[:final_stop_index]
+        return eg_idxes_for_calibration, precursor_df, fragments_df
+
+    #        self.eg_idxes_for_calibration = self.elution_group_order[:final_stop_index]
+    #        self.optimization_manager.fit({"classifier_version": self.fdr_manager.current_version})
+
+    def get_optimizers(self):
+        """Select appropriate optimizers. Targeted optimization is used if a valid target value (i.e. a number greater than 0) is specified in the config;
+        if a value less than or equal to 0 is supplied, automatic optimization is used.
```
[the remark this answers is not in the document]
odespard: DIA-NN works this way and so we are being consistent with that.

alphadia/workflow/optimization.py
odespard: I haven't done this because there is a good chance these methods will become parameter-specific as we improve the automatic calibration. I could change it to neaten up the code, but there is a good chance we would have to change them back again.

alphadia/workflow/optimization.py
odespard: As above: https://github.com/MannLabs/alphadia/pull/293#discussion_r1701422398.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,121 +282,198 @@ def get_batch_plan(self):
...
+        ]
+        automatic_optimizers = [
+            [optimizer]
+            for optimizer in optimizers
+            if isinstance(optimizer, optimization.AutomaticOptimizer)
+        ]
+
+        order_of_optimization = (
+            targeted_optimizers + automatic_optimizers
+            if any(targeted_optimizers)
```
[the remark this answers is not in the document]
odespard: It's not redundant: `any(targeted_optimizers) ` returns False for `targeted_optimizers=[[]]`, which is the case if there are no targeted optimizers in use (because of the way the optimizer lists are nested so all targeted optimizers run together). If the conditional is not present, this gives a result of [[], [automatic_optimizer_1], [automatic_optimizer_2], ...] in which the empty list causes an infinite loop later.

However, this isn't very clear from the code. I should at least write a comment to explain it, or think of a way to rewrite it.

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+            The dataframe used to update the parameter. This could be the precursor or fragment dataframe, depending on the search parameter being optimized.
+
+        Returns
+        -------
+        float
+            The proposed new value for the search parameter.
+
+
+        """
+        pass
```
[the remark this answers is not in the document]
odespard: Thanks. I didn't know that.

alphadia/workflow/optimization.py
mschwoer: if this chance is above 80% then it's fine .. otherwise I would always opt for implementing according to current requirements, because _cloudy the future is_ ;-)

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +484,151 @@ def calibration(self):
...
+        fragments_df : pd.DataFrame
+            Fragment dataframe.
+
+        Returns
+        -------
+        precursor_df_filtered : pd.DataFrame
+            Filtered precursor dataframe. Decoy precursors and those found at worse than 1% FDR are removed from the precursor dataframe.
+
+        fragments_df_filtered : pd.DataFrame
+            Filtered fragment dataframe. Retained fragments must have a correlation greater than 0.7 and belong to the top 5000 fragments sorted by correlation.
```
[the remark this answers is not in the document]
odespard: You are correct; the docstring is not right as it fails to deal with the case where there are fewer than 500 fragments with correlations above 0.7. I have corrected it.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,121 +282,198 @@ def get_batch_plan(self):
...
+            mobility_optimizer,
+        ]
+        targeted_optimizers = [
+            [
+                optimizer
+                for optimizer in optimizers
+                if isinstance(optimizer, optimization.TargetedOptimizer)
+            ]
+        ]
+        automatic_optimizers = [
```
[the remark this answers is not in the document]
odespard: We could. I could also rename the function to `get_ordered_optimizers` (and rename the variable to which its output is stored `ordered_optimizers`) to make the contract clearer.

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +484,151 @@ def calibration(self):
...
-        self.start_of_calibration()
-        for current_epoch in range(self.config["calibration"]["max_epochs"]):
-            if self.check_epoch_conditions():
-                pass
-            else:
-                break
+        # Get the order of optimization
+        order_of_optimization = self.get_optimizers()
+
+        self.reporter.log_string(
```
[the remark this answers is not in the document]
odespard: Maybe. If so, should I make this change throughout the pepticecentric.py file? 

The verbosity does change between "info" and "progress" quite a bit, so I don't know if a partial would be worth while.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,121 +282,198 @@ def get_batch_plan(self):
...
-                f"❌ {'rt_error':<15}: {self.com.rt_error:.4f} > {self.config['search']['target_rt_tolerance']}",
-                verbosity="info",
+                f"=== Classifier has been trained for {self.fdr_manager.current_version + 1} iteration(s); minimum number is {self.config['calibration']['optimization_lock_min_steps']} ===",
+                verbosity="progress",
+            )
+
+            if (
+                precursors_01FDR
+                > self.config["calibration"]["optimization_lock_target"]
+                and self.fdr_manager.current_version
```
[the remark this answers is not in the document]
odespard: It's called version because its primary purpose is to allow for retrieval of a previously-trained classifier when we backtrack to the optimum value of a parameter after the optimizer has converged. I think it's an intuitive name, but please do explain why it bugs you; I am sure it can probably be improved. 

I agree, however, that it's a confusing thing to use in this context because it is just acting as a synonym of current_step, which would be more intuitive to call. I have changed this line so that it uses current_step instead.

alphadia/workflow/peptidecentric.py
```diff
@@ -319,121 +282,198 @@ def get_batch_plan(self):
...
-        if self.com.ms2_error > self.config["search"]["target_ms2_tolerance"]:
             self.reporter.log_string(
-                f"❌ {'ms2_error':<15}: {self.com.ms2_error:.4f} > {self.config['search']['target_ms2_tolerance']}",
+                f"=== FDR correction performed with classifier version {self.optimization_manager.classifier_version} ===",
                 verbosity="info",
             )
-            continue_calibration = True
-        else:
+
+            precursors_01FDR = len(precursor_df[precursor_df["qval"] < 0.01])
```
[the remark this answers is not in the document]
odespard: I'm inclined to keep an explicit reference to the threshold since it is hard-coded. Would `num_precursors_at_01FDR` work?

alphadia/workflow/peptidecentric.py
```diff
@@ -444,89 +484,151 @@ def calibration(self):
...
-            for current_step, (start_index, stop_index) in enumerate(self.batch_plan):
-                self.start_of_step(current_step, start_index, stop_index)
+        for optimizers in order_of_optimization:
+            for optimizer in optimizers:
+                optimizer.step(precursor_df_filtered, fragments_df_filtered)
+        # End of first recalibration
 
-                eg_idxes = self.elution_group_order[start_index:stop_index]
+        # Start of optimization/recalibration loop
+        for optimizers in order_of_optimization:
```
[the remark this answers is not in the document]
odespard: Introducing a `first_recalibration_and_optimization()` method makes sense to me, as it is self-contained and somewhat distinct from the logic of the rest of calibration. But since a `recalibration_loop()` method would be more than half as long as the current `calibration()` method though, I'm not sure it would make things much clearer.

alphadia/workflow/optimization.py
```diff
@@ -0,0 +1,684 @@
...
+
+    @abstractmethod
+    def plot(self):
+        """
+        This method plots relevant information about optimization of the search parameter.
+
+        Notes
+        -----
+            This can be overwritten with pass if there is nothing of interest to plot.
```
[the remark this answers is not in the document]
odespard: Apparently the linter doesn't like this, giving this error:

`alphadia/workflow/optimization.py:56:5: B027 `BaseOptimizer.plot` is an empty method in an abstract base class, but has no abstract decorator
Found 1 error.`

alphadia/workflow/peptidecentric.py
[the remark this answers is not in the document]
odespard: I may revisit this but at the moment I have gone for the renaming solution.

alphadia/workflow/optimization.py
odespard: That could work, but I think it is OK as it is. Perhaps we can revisit this later.

alphadia/workflow/base.py
```diff
@@ -95,8 +95,22 @@ def load(
...
+            "column_type": "library",
+            "num_candidates": self.config["search_initial"]["initial_num_candidates"],
+            "classifier_version": -1,
+            "fwhm_rt": self.config["optimization_manager"]["fwhm_rt"],
+            "fwhm_mobility": self.config["optimization_manager"]["fwhm_mobility"],
+            "score_cutoff": self.config["optimization_manager"]["score_cutoff"],
+        }
         self._optimization_manager = manager.OptimizationManager(
-            self.config["optimization_manager"],
+            optimization_manager_config,
```
mschwoer, line 113: this seems to me like `OptimizationManager`-specific logic that is outside of `OptimizationManager` :-)
suggestion:
pass the `self.config` to the `OptimizationManager` and do this mapping during initialization. Also, I suggest separating between constant values (probable `fwhm_rt` does not change?) and dynamic values (e.g. classifier_version)

odespard: Yes, that is true. I'll change it. 

I think everything in the optimization manager should change – fwhm_rt for instance is updated during calibration.

PR #297, opened 2024-08-07
296 reduce alphadia output

mschwoer:
- add options to save library files (@GeorgWa is `general` the right place?)
- reduce tqdm output

Plus some small unrelated changes:
- fix issue when ouput folder parents do not exist
- make error logging more robust
- add debug instructions
- allow to skip neptune upload for convenient local testing


alphadia/outputtransform.py
```diff
@@ -365,7 +365,11 @@ def build(
...
         _ = self.build_lfq_tables(folder_list, psm_df=psm_df, save=True)
-        _ = self.build_library(base_spec_lib, psm_df=psm_df, save=True)
+        _ = self.build_library(
+            base_spec_lib,
+            psm_df=psm_df,
+            save=self.config["general"]["save_mbr_library"],
+        )
 
         if self.config["transfer_library"]["enabled"]:
             _ = self.build_transfer_library(folder_list, save=True)
```
mschwoer, line 375: should we make saving those guys also optional?

alphadia/constants/default.yaml
```diff
@@ -11,6 +11,9 @@ general:
   wsl: false
   mmap_detector_events: false
   use_gpu: true
+  # whether to save the (MBR) library to the output directory
+  save_library: True
```
GeorgWa, line 15: I think we could add that the first lib is the input library and the second is the output library.

alphadia/outputtransform.py
GeorgWa: Are they not already optional and only performed if transfer library enabled =True?

GeorgWa (approved):
I'm a bit opinionated about using self.config in the outputtransform  for consistency 😄
Otherwise it LGTM

alphadia/outputtransform.py
mschwoer: I was thinking  about the saving only ..

PR #298, opened 2024-08-08
fix test case data

mschwoer:
Fix apparent typo in test cases. Note that this will change numbers.


odespard:
It did strike me as odd that we analysed the same file twice as part of the test. I suppose one benefit of the current set-up is that it gives us some idea of how variable the analysis is when repeated on the same data.

odespard (commented):
It did strike me as odd that we analysed the same file twice as part of the test. I suppose one benefit of the current set-up is that it gives us some idea of how variable the analysis is when repeated on the same data.

GeorgWa (approved):
This explains why we always got 150k precursors :D
LGTM

PR #301, opened 2024-08-12
Calibrate optimization lock only during optimization

odespard:
Calibration occurs using only the precursors in the optimization lock. However, the calibration for the whole library is currently predicted at each step which adds considerably to the time taken for the computation. This changes the code such that the prediction is done only on the optimization lock, with the whole library only being predicted immediately prior to extraction.

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -1,6 +1,7 @@
 # native imports
 import logging
 import os
+import pickle
 from collections.abc import Iterator
 
 import directlfq.config as lfqconfig
@@ -653,9 +654,13 @@ class SearchPlanOutput:
         stat_df_list = []
         for folder in folder_list:
             raw_name = os.path.basename(folder)
+            optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
             stat_df_list.append(
                 _build_run_stat_df(
-                    raw_name, psm_df[psm_df["run"] == raw_name], all_channels
+                    raw_name,
+                    psm_df[psm_df["run"] == raw_name],
+                    optimization_manager_path,
+                    all_channels,
                 )
             )
 
@@ -820,7 +825,10 @@ class SearchPlanOutput:
 
 
 def _build_run_stat_df(
-    raw_name: str, run_df: pd.DataFrame, channels: list[int] | None = None
+    raw_name: str,
+    run_df: pd.DataFrame,
+    optimization_manager_path: str,
+    channels: list[int] | None = None,
 ):
     """Build stat dataframe for a single run.
 
@@ -865,6 +873,14 @@ def _build_run_stat_df(
 
         if "mobility_fwhm" in channel_df.columns:
             base_dict["fwhm_mobility"] = np.mean(channel_df["mobility_fwhm"])
+
+        with open(optimization_manager_path, "rb") as f:
+            optimization_manager = pickle.load(f)
+
+        base_dict["ms2_error"] = optimization_manager.ms2_error
+        base_dict["ms1_error"] = optimization_manager.ms1_error
+        base_dict["rt_error"] = optimization_manager.rt_error
+        base_dict["mobility_error"] = optimization_manager.mobility_error
 
         out_df.append(base_dict)
 
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -24,7 +24,7 @@ from alphadia.outputaccumulator import (
     TransferLearningAccumulator,
 )
 from alphadia.transferlearning.train import FinetuneManager
-from alphadia.workflow import manager
+from alphadia.workflow import manager, peptidecentric
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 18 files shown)


alphadia/workflow/peptidecentric.py
```diff
@@ -365,11 +365,46 @@ def get_optimization_lock(self):
...
+        optimization_lock_library_precursor_df = self.spectral_library._precursor_df[
+            self.spectral_library._precursor_df["elution_group_idx"].isin(
+                optimization_lock_eg_idxes
+            )
+        ]
+
+        optimization_lock_fragment_idxes = np.concatenate(
+            [
+                np.arange(row["flat_frag_start_idx"], row["flat_frag_stop_idx"])
+                for i, row in optimization_lock_library_precursor_df.iterrows()
```
mschwoer, line 379: (nit) `i` -> `_` (to mark it's not used)

alphadia/workflow/peptidecentric.py
```diff
@@ -365,11 +365,46 @@ def get_optimization_lock(self):
...
+            - optimization_lock_library_precursor_df["flat_frag_start_idx"]
+        )
+        optimization_lock_library_precursor_df["flat_frag_stop_idx"] = (
+            num_frags.cumsum()
+        )
+        optimization_lock_library_precursor_df["flat_frag_start_idx"] = (
+            optimization_lock_library_precursor_df[
+                "flat_frag_stop_idx"
+            ].shift(fill_value=0)
+        )
```
mschwoer, line 400: could all this new code be extracted into a dedicated method?
also: would be really cool to have a simple unit test for these kinds of index manipulations ;-)

alphadia/workflow/peptidecentric.py
```diff
@@ -757,7 +795,9 @@ def fdr_correction(self, features_df, df_fragments, version=-1):
             # neptune_run=self.neptune
         )
 
-    def extract_batch(self, batch_df, apply_cutoff=False):
+    def extract_batch(self, batch_df, batch_fragment_df=None, apply_cutoff=False):
```
mschwoer, line 798: rename `batch_df` => `batch_precursor_df` ?

mschwoer: (also in calling code)

mschwoer (approved):
LGTM

alphadia/outputtransform.py
```diff
@@ -653,9 +654,13 @@ def build_stat_df(
         stat_df_list = []
         for folder in folder_list:
             raw_name = os.path.basename(folder)
+            optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
```
GeorgWa, line 657: could we instead instantiate the `OptimizationManager(load_from_file=True)` here? We will most likely switch away from pickle at some point and then we only have the functionality for storing/loading within the `OptimizationManager`.

GeorgWa, line 657: Can you test / change it such that we report 0 if the `OptimizationManager` is not found and fail gracefully?

alphadia/workflow/manager.py
```diff
@@ -463,6 +463,18 @@ def __init__(
         self.reporter.log_string(f"Initializing {self.__class__.__name__}")
         self.reporter.log_event("initializing", {"name": f"{self.__class__.__name__}"})
 
+        initial_parameters = {
```
GeorgWa, line 466: makes sense to move them 👍🏻

alphadia/workflow/peptidecentric.py
GeorgWa: I agee :)

alphadia/workflow/peptidecentric.py
```diff
@@ -532,9 +567,12 @@ def calibration(self):
         )
 
         # Get the optimization lock
-        eg_idxes_for_calibration, precursor_df, fragments_df = (
-            self.get_optimization_lock()
-        )
+        (
+            self.optimization_lock_library_precursor_df,
```
GeorgWa, line 571: does this need to be a instance variable? I think we only need it within .calibration?

GeorgWa (commented):
Please have a look, should be good to merge soon.

alphadia/workflow/peptidecentric.py
odespard: I have split this into three methods in the next PR (https://github.com/MannLabs/alphadia/pull/304#issue-2463350503); I've also introduced other changes so this would be awkward to merge if it was changed here as well.

alphadia/outputtransform.py
odespard: That sounds like a good idea. I will change it.

odespard: I've added some code that checks if the path exists and gives a warning if it doesn't (but still continues). Is that OK?

alphadia/outputtransform.py
```diff
@@ -845,6 +847,8 @@ def _build_run_stat_df(
         Dataframe containing the statistics
 
     """
+    optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
```
GeorgWa, line 850: Can you use `OPTIMIZATION_MANAGER_PATH` here?

GeorgWa (approved):
LGTM

alphadia/workflow/peptidecentric.py
odespard: It's needed in recalibration as well. I could pass it as an argument but I thought it was easier to have it as an instance variable.

alphadia/outputtransform.py
odespard: Changed.

PR #302, opened 2024-08-13
add parameters to stat.tsv

odespard:
Uses stored optimization manager to include the tolerances used for search in the stat.tsv file. For targeted optimization this will be the same as those originally set but it is more useful for automatic optimization. I have also updated the tests to reflect the changes.

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -1,6 +1,7 @@
 # native imports
 import logging
 import os
+import pickle
 from collections.abc import Iterator
 
 import directlfq.config as lfqconfig
@@ -653,9 +654,13 @@ class SearchPlanOutput:
         stat_df_list = []
         for folder in folder_list:
             raw_name = os.path.basename(folder)
+            optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
             stat_df_list.append(
                 _build_run_stat_df(
-                    raw_name, psm_df[psm_df["run"] == raw_name], all_channels
+                    raw_name,
+                    psm_df[psm_df["run"] == raw_name],
+                    optimization_manager_path,
+                    all_channels,
                 )
             )
 
@@ -820,7 +825,10 @@ class SearchPlanOutput:
 
 
 def _build_run_stat_df(
-    raw_name: str, run_df: pd.DataFrame, channels: list[int] | None = None
+    raw_name: str,
+    run_df: pd.DataFrame,
+    optimization_manager_path: str,
+    channels: list[int] | None = None,
 ):
     """Build stat dataframe for a single run.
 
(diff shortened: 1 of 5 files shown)


alphadia/outputtransform.py
```diff
@@ -866,6 +874,14 @@ def _build_run_stat_df(
         if "mobility_fwhm" in channel_df.columns:
             base_dict["fwhm_mobility"] = np.mean(channel_df["mobility_fwhm"])
 
+        with open(optimization_manager_path, "rb") as f:
```
mschwoer, line 877: are we sure this file is always present or do we need to handle the error case?

alphadia/outputtransform.py
```diff
@@ -866,6 +874,14 @@ def _build_run_stat_df(
...
+            optimization_manager = pickle.load(f)
+
+        base_dict["ms2_error"] = optimization_manager.ms2_error
+        base_dict["ms1_error"] = optimization_manager.ms1_error
+        base_dict["rt_error"] = optimization_manager.rt_error
+        base_dict["mobility_error"] = optimization_manager.mobility_error
+
         out_df.append(base_dict)
```
mschwoer, line 885: `out_df` is a bit misleading as it's not a dataframe .. maybe rename to `channel_data`?

alphadia/workflow/manager.py
```diff
@@ -463,6 +463,18 @@ def __init__(
         self.reporter.log_string(f"Initializing {self.__class__.__name__}")
         self.reporter.log_event("initializing", {"name": f"{self.__class__.__name__}"})
 
+        initial_parameters = {
```
mschwoer, line 466: +1

tests/unit_tests/test_workflow.py
```diff
@@ -206,16 +206,26 @@ def test_calibration_manager_save_load():
     os.remove(temp_path)
 
 
-OPTIMIZATION_TEST_DATA = {
-    "fwhm_cycles": 5,
-    "fwhm_mobility": 0.01,
+OPTIMIZATION_CONFIG = {
```
mschwoer, line 209: (nit) I would always name test data such that is clear that it's just for tests, to avoid confusing it with real stuff .. `TEST_OPTIMIZATION_CONFIG` ?

tests/unit_tests/test_outputtransform.py
```diff
@@ -62,6 +75,11 @@ def test_output_transform():
         frag_df.to_parquet(os.path.join(raw_folder, "frag.parquet"), index=False)
         psm_df.to_parquet(os.path.join(raw_folder, "psm.parquet"), index=False)
 
+        optimization_manager = manager.OptimizationManager(
+            config, os.path.join(raw_folder, "optimization_manager.pkl")
```
mschwoer, line 79: is this pkl available in the tests?

mschwoer (approved):
LGTM!

alphadia/outputtransform.py
odespard: This is saved by default every time we fit something to the optimization manager, so it should always be present at this stage in the workflow.

tests/unit_tests/test_outputtransform.py
odespard: Yes, it's saved each time optimization_manager.fit() is called.

alphadia/outputtransform.py
odespard: That is true, but it's consistent throughout the file so I'm inclined not to change it just for this, and I'm reluctant to change it throughout on this pull request.

PR #303, opened 2024-08-13
extend transfer-learning to include mobility

mo-sameh:
Extended the transfer-learning to also include mobility finetuning.

### DIFF
--- a/alphadia/transferlearning/train.py
+++ b/alphadia/transferlearning/train.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import torch
 from alphabase.peptide.fragment import remove_unused_fragments
+from alphabase.peptide.mobility import ccs_to_mobility_for_df, mobility_to_ccs_for_df
 from peptdeep.model.charge import ChargeModelForModAASeq
 from peptdeep.model.model_interface import CallbackHandler, LR_SchedulerInterface
 from peptdeep.pretrained_models import ModelManager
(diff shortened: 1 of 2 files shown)


alphadia/transferlearning/train.py
```diff
@@ -896,3 +897,166 @@ def finetune_charge(self, psm_df: pd.DataFrame) -> pd.DataFrame:
...
+            The metric manager object.
+        data_split : str
+            The dataset label to test on. e.g. "validation", "train"
+        Returns
+        -------
+        bool
+            Whether to continue training or not based on the early stopping criteria.
+        """
+        continue_training = True
+        if epoch % self.settings["test_interval"] == 0:
```
mschwoer, line 930: this would save on indent:
```python
if epoch % self.settings["test_interval"] != 0:
    return continue_training

self.ccs_model.model.eval()
...
```

alphadia/transferlearning/train.py
```diff
@@ -896,3 +897,166 @@ def finetune_charge(self, psm_df: pd.DataFrame) -> pd.DataFrame:
...
+            pred = self.ccs_model.predict(test_df)
+
+            test_input = {
+                "predicted": pred["ccs_pred"].values,
+                "target": test_df["ccs"].values,
+            }
+            val_metrics = metric_accumulator.calculate_test_metric(
+                test_input, epoch, data_split=data_split, property_name="ccs"
+            )
+            if epoch != -1 and data_split == "validation":
```
mschwoer, line 942: does the check for `epoch != -1` make sense here?
if `epoch = -1`, we will be here only if `self.settings["test_interval"] = 1`

alphadia/transferlearning/train.py
```diff
@@ -896,3 +897,166 @@ def finetune_charge(self, psm_df: pd.DataFrame) -> pd.DataFrame:
...
+
+        # Shuffle the psm_df and split it into train and test
+        train_df = psm_df.sample(frac=self.settings["train_fraction"])
+        val_df = psm_df.drop(train_df.index).sample(
+            frac=self.settings["validation_fraction"]
+            / (1 - self.settings["train_fraction"])
+        )
+        test_df = psm_df.drop(train_df.index).drop(val_df.index)
+
+        # Create a test metric manager
```
mschwoer, line 1010: (nit) I think the code is very very clear and needs no further explanation ;-)

alphadia/transferlearning/train.py
```diff
@@ -896,3 +897,166 @@ def finetune_charge(self, psm_df: pd.DataFrame) -> pd.DataFrame:
...
+        self.ccs_model.model.train()
+        self.ccs_model.train(
+            train_df,
+            batch_size=self.settings["batch_size"],
+            epoch=self.settings["epochs"],
+            warmup_epoch=self.settings["warmup_epochs"],
+            lr=settings["max_lr"],
+        )
+
+        self._test_ccs(
```
mschwoer, line 1056: we are not using the return value?

alphadia/transferlearning/train.py
```diff
@@ -896,3 +897,166 @@ def finetune_charge(self, psm_df: pd.DataFrame) -> pd.DataFrame:
...
+            else:
+                logger.progress(
+                    f" CCS model tested on {data_split} dataset with the following metrics:"
+                )
+                for i in range(len(val_metrics)):
+                    logger.progress(
+                        f" {val_metrics['metric_name'].values[i]:<30}: {val_metrics['value'].values[i]:.4f}"
+                    )
+
+            self.ccs_model.model.train()
```
mschwoer, line 974: from the name and docstring of this method I would assume this method just tests, but it also trains?

mschwoer (approved):
LGTM!

alphadia/transferlearning/train.py
mschwoer: thanks for the clarification!

GeorgWa (approved):
LGTM

PR #304, opened 2024-08-13
Dynamic optimization lock

odespard:
Reduce the size of the optimization lock as detection of precursors increases.

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -40,7 +40,6 @@ class BaseOptimizer(ABC):
         precursors_df: pd.DataFrame,
         fragments_df: pd.DataFrame,
         current_step: int,
-        record_step: bool = True,
     ):
         """This method evaluates the progress of the optimization, and either concludes the optimization if it has converged or continues the optimization if it has not.
         This method includes the update rule for the optimization.
@@ -57,8 +56,6 @@ class BaseOptimizer(ABC):
         current_step: int
             The current step in the optimization process. By default it is set to -1, which prevents the optimizer from converging unless min_steps has been set to 0.
 
-        record_step: bool
-            Whether to record the current step in the optimization process. If False, the step will not be recorded in the history dataframe. If the implementation of the optimizer does not record the history of the optimization, this parameter can be ignored.
         """
 
         pass
@@ -97,7 +94,6 @@ class AutomaticOptimizer(BaseOptimizer):
         precursors_df: pd.DataFrame,
         fragments_df: pd.DataFrame,
         current_step: int = -1,
-        record_step: bool = True,
     ):
         """See base class. The feature is used to track the progres of the optimization (stored in .feature) and determine whether it has converged."""
         if self.has_converged:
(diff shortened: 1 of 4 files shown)


alphadia/workflow/optimization.py
```diff
@@ -95,24 +96,42 @@ def step(
...
+                    precursors_df.sort_values(by="precursor_idx")
+                    .iloc[
+                        2
+                        * self.workflow.config["calibration"][
+                            "optimization_lock_target"
+                        ]
+                    ]
+                    .precursor_idx
+                )
+            except IndexError:
```
mschwoer, line 112: I know that "favor 'try' over 'upfront check'" is the pythonic way, but I feel in this case it would increase readability if the check was explictly done ..
as far as I got it
```python
num_targets = 2* self.workflow.config["calibration"][
                            "optimization_lock_target"
                        ]
if len(precursors_df) >=  num_targets:
...
else
...
```

alphadia/workflow/optimization.py
```diff
@@ -95,24 +96,42 @@ def step(
                 verbosity="progress",
             )
             return
+        if record_step:
+            try:
+                # Takes the target number of precursors if at least the target is found,
+                precursor_cutoff_idx = (
+                    precursors_df.sort_values(by="precursor_idx")
+                    .iloc[
+                        2
```
mschwoer, line 105: sure you want to pick only one index? or is there a `:` missing ;-)

alphadia/workflow/optimization.py
```diff
@@ -335,6 +358,7 @@ def step(
         precursors_df: pd.DataFrame,
         fragments_df: pd.DataFrame,
         current_step: int = -1,
+        record_step: bool = False,
```
mschwoer, line 361: please add a docstring explaining this parameter

alphadia/workflow/optimization.py
```diff
@@ -87,6 +87,7 @@ def step(
         precursors_df: pd.DataFrame,
         fragments_df: pd.DataFrame,
         current_step: int = -1,
+        record_step: bool = True,
```
mschwoer, line 90: please add a docstring explaining this parameter

alphadia/workflow/peptidecentric.py
```diff
@@ -366,18 +366,33 @@ def get_optimization_lock(self):
...
 
         optimization_lock_eg_idxes = self.elution_group_order[:final_stop_index]
-
         optimization_lock_library_precursor_df = self.spectral_library._precursor_df[
             self.spectral_library._precursor_df["elution_group_idx"].isin(
                 optimization_lock_eg_idxes
             )
         ]
 
+        randomized_idx = np.random.permutation(
```
mschwoer, line 375: in order to increase reproducibility: should we fix the seed? or at least enable fixing it (e.g. via config)?

alphadia/workflow/peptidecentric.py
```diff
@@ -366,18 +366,33 @@ def get_optimization_lock(self):
...
+            optimization_lock_library_precursor_df["precursor_idx"].values
+        )
+        self.randomized_idx_mapping = dict(
+            zip(
+                optimization_lock_library_precursor_df["precursor_idx"].values,
+                randomized_idx,
+            )
```
mschwoer, line 382: please add `strict=True` to `zip`

mschwoer: (also below)

alphadia/workflow/peptidecentric.py
```diff
@@ -537,7 +553,53 @@ def first_recalibration_and_optimization(
...
+            "precursor_idx"
+        ].map(self.randomized_idx_mapping)
+        try:
+            # Takes the target number of precursors if at least the target is found
+            self.precursor_cutoff_idx = (
+                precursor_df_filtered.sort_values(by="precursor_idx")
+                .iloc[self.config["calibration"]["optimization_lock_target"]]
+                .precursor_idx
+            )
+        except IndexError:
```
mschwoer, line 570: same as above: I would prefer an explicit check

alphadia/workflow/peptidecentric.py
```diff
@@ -537,7 +553,53 @@ def first_recalibration_and_optimization(
...
+
+    def set_reduced_optimization_lock(self):
+        self.reduced_optimization_lock_library_precursor_df = (
+            self.optimization_lock_library_precursor_df[
+                self.optimization_lock_library_precursor_df["precursor_idx"]
+                < self.precursor_cutoff_idx
+            ]
+        )
+
+        fragment_cutoff_idx = self.reduced_optimization_lock_library_precursor_df[
```
mschwoer, line 589: this is a reeeeally long variable name ..

think about introucing `optlock` as subsitute for `optimization_lock` throughout all the codebase (if you choose to do so, please do the renaming in a dedicated PR)

then this could become `red_optlock_lib_precursor_df`

alphadia/workflow/optimization.py
odespard: In this case, we just want the correct cutoff index, so only one index is wanted.

mschwoer: all good, maybe introduce `cutoff_index = 2 * .. ` then and add a little explanation (e.g. on the factor of 2 ;-))

PR #306, opened 2024-08-14
timing manager

odespard:
For tracking performance of different optimization methods, it will be useful to know how long it takes to perform both optimization and extraction for a given file. In this revision, this information is recorded by the timing manager and returned to the user in the stat.tsv file.

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -1,7 +1,6 @@
 # native imports
 import logging
 import os
-import pickle
 from collections.abc import Iterator
 
 import directlfq.config as lfqconfig
@@ -25,6 +24,7 @@ from alphadia.outputaccumulator import (
     TransferLearningAccumulator,
 )
 from alphadia.transferlearning.train import FinetuneManager
+from alphadia.workflow import manager
 
 logger = logging.getLogger()
 
@@ -657,12 +657,11 @@ class SearchPlanOutput:
         stat_df_list = []
         for folder in folder_list:
             raw_name = os.path.basename(folder)
-            optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
             stat_df_list.append(
                 _build_run_stat_df(
+                    folder,
                     raw_name,
                     psm_df[psm_df["run"] == raw_name],
-                    optimization_manager_path,
                     all_channels,
                 )
             )
@@ -820,15 +819,18 @@ class SearchPlanOutput:
 
 
 def _build_run_stat_df(
+    folder: str,
     raw_name: str,
     run_df: pd.DataFrame,
-    optimization_manager_path: str,
     channels: list[int] | None = None,
 ):
     """Build stat dataframe for a single run.
 
     Parameters
     ----------
+
+    folder: str
+        Directory containing the raw file and the managers
 
     raw_name: str
         Name of the raw file
@@ -845,6 +847,8 @@ def _build_run_stat_df(
         Dataframe containing the statistics
 
     """
+    optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
+    timing_manager_path = os.path.join(folder, "timing_manager.pkl")
 
     if channels is None:
         channels = [0]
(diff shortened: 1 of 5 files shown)


alphadia/outputtransform.py
```diff
@@ -820,16 +819,19 @@ def build_library(
...
-    optimization_manager_path: str,
     channels: list[int] | None = None,
 ):
     """Build stat dataframe for a single run.
 
     Parameters
     ----------
 
+    folder: str
+        Directory containing the raw file and the managers
```
mschwoer, line 833: is this a folder name? a relative path? an absolute path? please add this information to the docstring and conaider changing also the name (e.g.  to`output_folder_path`)

alphadia/outputtransform.py
```diff
@@ -845,6 +847,8 @@ def _build_run_stat_df(
         Dataframe containing the statistics
 
     """
+    optimization_manager_path = os.path.join(folder, "optimization_manager.pkl")
```
mschwoer, line 850: could we move this logic to the Managers?

cf. also this comment: https://github.com/MannLabs/alphadia/pull/301#discussion_r1717115466

alphadia/workflow/manager.py
```diff
@@ -750,3 +752,43 @@ def fit(self, update_dict):
 def column_hash(columns):
     columns.sort()
     return xxhash.xxh64_hexdigest("".join(columns))
+
+
+class TimingManager(BaseManager):
+    def __init__(
```
mschwoer, line 758: please add docstrings

alphadia/workflow/manager.py
```diff
@@ -750,3 +752,43 @@ def fit(self, update_dict):
...
+        path: None | str = None,
+        load_from_file: bool = True,
+        **kwargs,
+    ):
+        super().__init__(path=path, load_from_file=load_from_file, **kwargs)
+        self.reporter.log_string(f"Initializing {self.__class__.__name__}")
+        self.reporter.log_event("initializing", {"name": f"{self.__class__.__name__}"})
+
+        if not self.is_loaded_from_file:
+            self.__dict__.update(
```
mschwoer, line 769: I know the other managers use the `__dict__` idiom, but could we start trying to avoid it?
(unless there's a good reason, @GeorgWa ?)

cf. https://stackoverflow.com/a/60105123

alphadia/workflow/manager.py
```diff
@@ -750,3 +752,43 @@ def fit(self, update_dict):
...
+                        "duration": None,
+                    },
+                }
+            )
+
+    def start(self, workflow_stage: str):
+        self.__dict__.update({workflow_stage: {"start": pd.Timestamp.now()}})
+        self.save()
+
+    def end(self, workflow_stage: str):
```
mschwoer, line 788: `stop`?
or `set_start_time`, `set_end_time`?

alphadia/workflow/manager.py
mschwoer: from that I see, this could be a dictionary stored in an instance variable

alphadia/workflow/peptidecentric.py
```diff
@@ -894,6 +897,8 @@ def extraction(self):
 
         self.log_precursor_df(precursor_df)
 
+        self.timing_manager.end("extraction")
```
mschwoer, line 900: if `self.timing_manager.start("extraction")` was called again, this would overwrite the previously recorded timeing, right? is this intended? or should we change it such that it accumulates?

alphadia/outputtransform.py
```diff
@@ -869,13 +873,33 @@ def _build_run_stat_df(
...
+            base_dict["mobility_error"] = np.nan
+
+        if os.path.exists(timing_manager_path):
+            timing_manager = manager.TimingManager(path=timing_manager_path)
 
-        base_dict["ms2_error"] = optimization_manager.ms2_error
-        base_dict["ms1_error"] = optimization_manager.ms1_error
-        base_dict["rt_error"] = optimization_manager.rt_error
-        base_dict["mobility_error"] = optimization_manager.mobility_error
+            base_dict["optimization_duration"] = timing_manager.optimization["duration"]
```
mschwoer, line 896: if `TimingManager` had this info stored in an actual dictionary (e.g. `timings`), you could do smth like

```python
for key in timing_manager.timings:
  base_dict[f"duration_{key}"] = timing_manager.timings[key]
```
thus avoiding an explicit coupling of this to the individual timings
(cf also the comment below)

alphadia/workflow/peptidecentric.py
odespard: I suppose it would. Why would we want to start the same timing twice though (unless we were trying to time iterations of a loop or something like that, but then I think we would need to make the whole logic of the timing manager considerably more complex)?

PR #307, opened 2024-08-15
FIX bug in timstof transpose function

GeorgWa:
Fixed bug where the transposed matrix would approximate the shape incorrectly due to missing values.
Changed testcase to catch this in the future.


odespard (approved):
Excellent. I hope this stops the segmentation faults.

PR #309, opened 2024-08-16
New dynamic optimization lock

odespard:
Allows optimization lock increase and decrease in size, which prevents crashing if the number of precursors detected decreases markedly. Also encapsulates optimization lock-related functions in a new class.

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -297,6 +297,7 @@ class SearchPlanOutput:
     PSM_INPUT = "psm"
     PRECURSOR_OUTPUT = "precursors"
     STAT_OUTPUT = "stat"
+    INTERNAL_OUTPUT = "internal"
     PG_OUTPUT = "protein_groups"
     LIBRARY_OUTPUT = "speclib.mbr"
     TRANSFER_OUTPUT = "speclib.transfer"
@@ -365,6 +366,7 @@ class SearchPlanOutput:
             folder_list, save=False, base_spec_lib=base_spec_lib
         )
         _ = self.build_stat_df(folder_list, psm_df=psm_df, save=True)
+        _ = self.build_internal_df(folder_list, save=True)
         _ = self.build_lfq_tables(folder_list, psm_df=psm_df, save=True)
         _ = self.build_library(
             base_spec_lib,
@@ -678,6 +680,50 @@ class SearchPlanOutput:
 
         return stat_df
 
+    def build_internal_df(
+        self,
+        folder_list: list[str],
+        save: bool = True,
+    ):
+        """Build internal data table from a list of seach outputs
+
+        Parameters
+        ----------
+
+        folder_list: List[str]
+            List of folders containing the search outputs
+
+        save: bool
+            Save the precursor table to disk
+
+        Returns
+        -------
+
+        stat_df: pd.DataFrame
+            Precursor table
+        """
+        logger.progress("Building internal statistics")
+
+        internal_df_list = []
+        for folder in folder_list:
+            internal_df_list.append(
+                _build_run_internal_df(
+                    folder,
+                )
+            )
+
+        internal_df = pd.concat(internal_df_list)
+
+        if save:
+            logger.info("Writing internal output to disk")
+            write_df(
+                internal_df,
+                os.path.join(self.output_folder, self.INTERNAL_OUTPUT),
+                file_format="tsv",
+            )
+
+        return internal_df
+
     def build_lfq_tables(
         self,
         folder_list: list[str],
@@ -850,9 +896,6 @@ def _build_run_stat_df(
     optimization_manager_path = os.path.join(
         folder, peptidecentric.PeptideCentricWorkflow.OPTIMIZATION_MANAGER_PATH
     )
-    timing_manager_path = os.path.join(
-        folder, peptidecentric.PeptideCentricWorkflow.TIMING_MANAGER_PATH
-    )
 
     if channels is None:
         channels = [0]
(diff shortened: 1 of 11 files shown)


alphadia/workflow/optimization.py
```diff
@@ -579,3 +575,152 @@ def __init__(
         self.estimator_group_name = "precursor"
         self.estimator_name = "mobility"
         super().__init__(initial_parameter, target_parameter, workflow, reporter)
+
+
+class OptimizationLock:
```
mschwoer, line 580: would be really cool to fully cover this new class with unit tests

alphadia/workflow/peptidecentric.py
```diff
@@ -601,54 +425,64 @@ def calibration(self):
...
 
                 self.reporter.log_string(
                     f"=== FDR correction performed with classifier version {self.optimization_manager.classifier_version} ===",
                     verbosity="info",
                 )
 
-                self.log_precursor_df(precursor_df)
+                self.log_precursor_df(self.optlock.precursor_df)
+
+                if self.optlock.reached_target:
```
mschwoer, line 448: so we need to reach the target twice? think about more defined names here e.g. `reached_n_precursors_target`, `reached_optimization_target` .. ?

alphadia/workflow/optimization.py
```diff
@@ -579,3 +575,152 @@ def __init__(
...
+        )
+
+        self.count = len(self.precursor_df[self.precursor_df["qval"] < 0.01])
+
+        if self.count < self.target_count:
+            self.reached_target = False
+        else:
+            self.reached_target = True
+
+    def adjust_upwards(self):
```
GeorgWa, line 669: Do `adjust_upwards` and `reset_downwards` do the same just in opposite directions? 
If so, could we rename them to make this more clear?

alphadia/workflow/optimization.py
```diff
@@ -579,3 +575,152 @@ def __init__(
...
+
+        Parameters
+        ----------
+        workflow: peptidecentric.PeptideCentricWorkflow
+            The workflow object, which includes the spectral library and the calibration and optimization managers.
+
+        """
+        self.workflow = workflow
+
+        self.reached_target_before = False
```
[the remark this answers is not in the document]
odespard: I have changed the interface so it takes only specific parts of the workflow object.

alphadia/outputtransform.py
```diff
@@ -678,6 +680,50 @@ def build_stat_df(
 
         return stat_df
 
+    def build_internal_df(
+        self,
+        folder_list: list[str],
```
mschwoer, line 685: (nit) `search_output_folders`

alphadia/outputtransform.py
```diff
@@ -894,20 +937,45 @@ def _build_run_stat_df(
...
-            base_dict["extraction_duration"] = timing_manager.extraction["duration"]
+    return pd.DataFrame(out_df)
 
-        else:
-            logger.warning(f"Error reading timing manager for {raw_name}")
-            base_dict["optimization_duration"] = np.nan
-            base_dict["extraction_duration"] = np.nan
 
-        out_df.append(base_dict)
+def _build_run_internal_df(
```
mschwoer, line 945: this could have been a new PR ;-)

alphadia/workflow/manager.py
```diff
@@ -761,34 +760,22 @@ def __init__(
...
-                        "duration": None,
-                    },
-                }
-            )
+            self.timings = {}
 
-    def start(self, workflow_stage: str):
-        self.__dict__.update({workflow_stage: {"start": pd.Timestamp.now()}})
+    def set_start_time(self, workflow_stage: str):
+        """Starts the timer for a portion of the workflow and stores it in the timings attribute under the name given by the parameter workflow_stage"""
```
mschwoer, line 771: (nit) does this really start a timer? ;-)

alphadia/workflow/peptidecentric.py
```diff
@@ -521,24 +370,17 @@ def first_recalibration_and_optimization(
...
-        self.reporter.log_string(
-            "=== Performing initial optimization on extracted data. ===",
-            verbosity="info",
-        )
-
-        for optimizers in ordered_optimizers:
-            for optimizer in optimizers:
-                optimizer.step(precursor_df_filtered, fragments_df_filtered)
-
     def calibration(self):
```
mschwoer, line 384: should we rename this to `search_parameter_optimization`?

alphadia/workflow/peptidecentric.py
```diff
@@ -557,40 +399,22 @@ def calibration(self):
                 verbosity="progress",
             )
             return
-        self.timing_manager.start("optimization")
+        self.timing_manager.set_start_time("optimization")
```
mschwoer, line 402: shall we move the timing logic one level up to the calling code? (also for "extraction")

alphadia/outputtransform.py
```diff
@@ -894,20 +937,45 @@ def _build_run_stat_df(
...
+        folder_path, peptidecentric.PeptideCentricWorkflow.TIMING_MANAGER_PATH
+    )
+    raw_name = os.path.basename(folder_path)
+
+    internal_dict = {}
+
+    if os.path.exists(timing_manager_path):
+        timing_manager = manager.TimingManager(path=timing_manager_path)
+        for key in timing_manager.timings:
+            internal_dict[f"{key}_duration"] = [timing_manager.timings[key]["duration"]]
```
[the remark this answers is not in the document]
mschwoer: or `time_elapsed_{key}`?

alphadia/workflow/optimization.py
```diff
@@ -579,3 +576,164 @@ def __init__(
...
+
+        Parameters
+        ----------
+        library: alphabase.spectral_library.flat.SpecLibFlat
+            The library object from the PeptideCentricWorkflow object, which includes the precursor and fragment library dataframes.
+
+        config: dict
+            The configuration dictionary from the PeptideCentricWorkflow object.
+        """
+        self.library = library
```
mschwoer, line 593: please prefix all private instance variables and methods with `_`

mschwoer: (calls from a test class don't count ;-))

alphadia/workflow/optimization.py
```diff
@@ -579,3 +576,164 @@ def __init__(
...
+
+    def decrease_batch_idx(self):
+        """If the optimization lock contains enough precursors at 1% FDR, checks if enough precursors can be obtained using a smaller library and updates the library attribute accordingly.
+        If not, the same library is used as before.
+        This is done by taking the smallest step in the batch plan which gives more precursors than the target number of precursors.
+        """
+
+        batch_plan_diff = np.array(
+            [
+                stop_at_given_idx - self.stop_idx * self.target_count / self.count
```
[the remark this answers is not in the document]
odespard: Yes, it normally is a non-integer, but this doesn't pose a problem since this is just a way to find the smallest positive difference (after which we then take the index of this position).

alphadia/outputtransform.py
odespard: `folder list` is used consistently throughout the file. If I do change it, should I change it everywhere (and if so should this be a separate, small PR)?

alphadia/workflow/peptidecentric.py
odespard: Or just `optimization`?

mschwoer: would be less specific (there might be other optimizations?), but also fine ..

odespard: I suppose so; let's go for the longer name then.

alphadia/planning.py
```diff
@@ -329,6 +329,7 @@ def run(
                 workflow.timing_manager.set_start_time("extraction")
                 psm_df, frag_df = workflow.extraction()
                 workflow.timing_manager.set_end_time("extraction")
+                workflow.timing_manager.save()
```
odespard, line 332: I have moved the saving of the timing manager to this point (rather than saving it with each call) because it helps with error handling. At the moment we check if the `timing_manager.pkl` file exists and fail gracefully (by moving to the next file instead of crashing) if it does not. However, the search will crash if the ``timing_manager.pkl`` file exists but is incomplete because an error arose after it was first saved but before it was complete. This should stop such errors arising in the future.

GeorgWa (approved):
LGTM

PR #311, opened 2024-08-19
BUG max-lr not used correctly

GeorgWa:
The maximum learning rate was not correctly read from the settings.

### DIFF
--- a/alphadia/transferlearning/train.py
+++ b/alphadia/transferlearning/train.py
@@ -547,7 +547,7 @@ class FinetuneManager(ModelManager):
             epoch=self.settings["epochs"],
             batch_size=self.settings["batch_size"],
             warmup_epoch=self.settings["warmup_epochs"],
-            lr=settings["max_lr"],
+            lr=self.settings["max_lr"],
         )
 
         self._test_ms2(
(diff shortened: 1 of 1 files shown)


alphadia/transferlearning/train.py
```diff
@@ -547,7 +547,7 @@ def finetune_ms2(
             epoch=self.settings["epochs"],
             batch_size=self.settings["batch_size"],
             warmup_epoch=self.settings["warmup_epochs"],
-            lr=settings["max_lr"],
+            lr=self.settings["max_lr"],
```
mschwoer, line 550: what's the role of this `settings` object that this code was accidentially referring to?
could it be removed?
in code, it's referenced only here:
```python
self.reduce_lr_on_plateau = torch.optim.lr_scheduler.ReduceLROnPlateau(
            optimizer,
            mode="min",
            patience=settings["lr_patience"],
            factor=0.5,
        )
```
(and in one notebook)

if it is to be kept: should we rename it? add some docstring?

GeorgWa: it shouldn't be referenced anywhere :D good point

GeorgWa: @mo-sameh Can we get rid of the settings object and use kwargs with defaults?

mschwoer: now as far as I see the only usage left is in `finetuning.ipynb` .. we could just move it there?

GeorgWa: Will be completely moved to kwargs 👍 by mo

mschwoer (approved):

PR #312, opened 2024-08-19
FEAT enable selection of alternative peptdeep models

GeorgWa:
Allows selection of alternative pretrained models shipped with peptdeep

### DIFF
--- a/alphadia/libtransform.py
+++ b/alphadia/libtransform.py
@@ -281,7 +281,8 @@ class PeptDeepPrediction(ProcessingStep):
             Path to a folder containing PeptDeep models. If not provided, the default models will be used.
 
         peptdeep_model_type : str, optional
-            Use other peptdeep models provided by the peptdeep model manager. Default is None.
+            Use other peptdeep models provided by the peptdeep model manager.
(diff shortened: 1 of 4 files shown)


alphadia/libtransform.py
```diff
@@ -313,6 +319,11 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
         device = utils.get_torch_device(self.use_gpu)
 
         model_mgr = ModelManager(device=device)
+
+        if self.peptdeep_model_type is not None:
```
mschwoer, line 323: mb add a comment here that this overwrites the `generic` model loaded in `ModelManager`

alphadia/libtransform.py
```diff
@@ -279,6 +280,10 @@ def __init__(
         peptdeep_model_path : str, optional
             Path to a folder containing PeptDeep models. If not provided, the default models will be used.
 
+        peptdeep_model_type : str, optional
+            Use other peptdeep models provided by the peptdeep model manager. Default is None.
```
mschwoer, line 284: `Default is None, which means the peptdeep default model ("generic") is being used.

mschwoer (approved):

PR #314, opened 2024-08-19
BUG multiplexing fails due to missing column

GeorgWa:


alphadia/plexscoring.py
```diff
@@ -1722,8 +1722,10 @@ def collect_candidates(
             "frame_center",
             "frame_start",
             "frame_stop",
-            "score",
         ]
+
+        candidate_df_columns += ["score"] if "score" in candidates_df.columns else []
```
mschwoer, line 1727: the fact that this fix is a one liner is a little hint that there might be some obscurity and hidden contract here ;-)

mschwoer: mb comment why `score` could be missing?

mschwoer (approved):

alphadia/plexscoring.py
GeorgWa: Yes, indeed :D we would need some form of validated dataframes for this

PR #316, opened 2024-08-20
New dynamic lock with RT proportion

odespard:
Take a proportion of the total gradient length given an initial RT tolerance less than 1.

### DIFF
--- a/alphadia/workflow/base.py
+++ b/alphadia/workflow/base.py
@@ -102,7 +102,9 @@ class WorkflowBase:
             self._calibration_manager.disable_mobility_calibration()
 
         # initialize the optimization manager
-
+        self.config["optimization_manager"]["gradient_length"] = (
+            self.dia_data.rt_values.max()
+        )
         self._optimization_manager = manager.OptimizationManager(
             self.config,
             path=os.path.join(self.path, self.OPTIMIZATION_MANAGER_PATH),
--- a/alphadia/workflow/base.py
+++ b/alphadia/workflow/base.py
@@ -102,10 +102,8 @@ class WorkflowBase:
             self._calibration_manager.disable_mobility_calibration()
 
         # initialize the optimization manager
-        self.config["optimization_manager"]["gradient_length"] = (
-            self.dia_data.rt_values.max()
-        )
         self._optimization_manager = manager.OptimizationManager(
+            self.dia_data.rt_values.max(),
             self.config,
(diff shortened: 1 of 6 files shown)


alphadia/workflow/base.py
```diff
@@ -102,7 +102,9 @@ def load(
             self._calibration_manager.disable_mobility_calibration()
 
         # initialize the optimization manager
-
+        self.config["optimization_manager"]["gradient_length"] = (
+            self.dia_data.rt_values.max()
```
GeorgWa, line 106: I think we can avoid to edit the config dynamically.

Maybe we can just use workflow.dia_data.rt_values within the respective optimizer.

tests/unit_tests/test_workflow.py
```diff
@@ -218,6 +218,7 @@ def test_calibration_manager_save_load():
         "fwhm_rt": 5,
         "fwhm_mobility": 0.01,
         "score_cutoff": 50,
+        "gradient_length": 1000,
```
GeorgWa, line 221: I think we can get away without modifying the config during processing to avoid chaos :) so this shouldn't be necessary.

GeorgWa (approved):
Should be good to merge without the config modification :)

alphadia/workflow/base.py
odespard: I did it this way because the config is already dynamic (in the sense that the output folder field is only created and assigned during the planning phase), and I thought that perhaps we might want to be able to specify the gradient length in the config. But I am very happy to change it.

alphadia/workflow/manager.py
```diff
@@ -467,7 +467,10 @@ def __init__(
             initial_parameters = {
                 "ms1_error": config["search_initial"]["initial_ms1_tolerance"],
                 "ms2_error": config["search_initial"]["initial_ms2_tolerance"],
-                "rt_error": config["search_initial"]["initial_rt_tolerance"],
+                "rt_error": config["search_initial"]["initial_rt_tolerance"]
```
mschwoer, line 470: please introduce a variable `rt_error` for this nontrivial logic

alphadia/workflow/base.py
mschwoer: I would also vote for the config being static (we could even enforce this using `MappingProxyType`: https://stackoverflow.com/a/72371333)

PR #318, opened 2024-08-21
Make optimization configurable

odespard:
Add a new ``optimization`` field in the config which can be used to specify the update rule parameters and the order of optimization.

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -190,7 +190,15 @@ search_output:
   file_format: "tsv"
 
 optimization:
+  # The order in which to perform optimization. Should be a list of lists of parameter names
+  # Example: [['ms1_error', 'ms2_error', 'rt_error', 'mobility_error']] means that all parameters are optimized simultaneously.
+  # Example: [["ms2_error"], ["rt_error"], ["ms1_error"], ["mobility_error"]] means that the parameters are optimized sequentially in the order given.
+  # Example: [["rt_error"], ["ms1_error", "ms2_error"]] means that first rt_error is optimized, then ms1_error and ms2_error are optimized simultaneously, and mobility_error is not optimized at all.
   order_of_optimization: null
+
+  # Parameters for the update rule for each parameter:
+  #   - update_interval: the percentile interval to use (as a decimal)
+  #   - update_factor: the factor by which to multiply the result from the percentile interval to get the new parameter value for the next round of search
   ms2_error:
       update_interval: 0.95
       update_factor: 1.0
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -192,17 +192,17 @@ search_output:
 optimization:
   order_of_optimization: null
   ms2_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   ms1_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   mobility_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   rt_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
 
 # configuration for the optimization manager
 # initial parameters, will be optimized
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -191,14 +191,18 @@ search_output:
 
 optimization:
   order_of_optimization: null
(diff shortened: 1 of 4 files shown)


alphadia/constants/default.yaml
```diff
@@ -189,8 +189,23 @@ search_output:
   # can be either "parquet" or "tsv"
   file_format: "tsv"
 
+optimization:
+  order_of_optimization: null
```
mschwoer, line 193: please comment heavily here on what these parameters mean.

for `order_of_optimization`, state what is expected (list? integers? names?) and give an example (or set the default here)

alphadia/constants/default.yaml
```diff
@@ -189,8 +189,23 @@ search_output:
...
+  ms2_error:
+      update_interval: 0.95
```
mschwoer, line 195: sure these are all correct? would have expected 0.99 and 1.1 here for some ..

alphadia/workflow/optimization.py
```diff
@@ -508,9 +523,9 @@ def _propose_new_parameter(self, df: pd.DataFrame):
 
         """
 
-        return 1.1 * self.workflow.calibration_manager.get_estimator(
+        return self.update_factor * self.workflow.calibration_manager.get_estimator(
```
mschwoer, line 526: now that we parametrized it, coming back to an earlier suggestion: could we have only one implementation of this?
(this would be a dedicated PR)

alphadia/workflow/peptidecentric.py
```diff
@@ -323,32 +323,51 @@ def get_ordered_optimizers(self):
...
-        optimizers = [
-            ms2_optimizer,
-            rt_optimizer,
-            ms1_optimizer,
-            mobility_optimizer,
-        ]
-        targeted_optimizers = [
-            [
-                optimizer
+        if self.config["optimization"]["order_of_optimization"] is None:
```
mschwoer, line 326: the `if` branch could be removed if the `default.yaml` provided a default instead of `None`

mschwoer (commented):
some suggestions for simplifications

alphadia/constants/default.yaml
odespard: It depends on whether we want to allow distinct update rules for targeted and automatic calibration. I decided not to (since any given run will only use one of these) and, since the emphasis so far has been on preventing any performance degradation for the targeted case, I thought it best to use the default values for this case in the YAML.

alphadia/workflow/peptidecentric.py
odespard: It couldn't be removed while maintaining precisely the same functionality, which is to ensure that targeted optimizers always come before automatic optimizers. Whether this functionality is worth maintaining is a separate question, but I thought it best to leave it in place for the time being as it seems a reasonable default.

alphadia/workflow/optimization.py
odespard: We could, and I suppose we could always overwrite the default method in a child class if needed. I'll make do this soon.

alphadia/constants/default.yaml
odespard: I'll add comments. I am inclined not to set a default for the reasons set out in https://github.com/MannLabs/alphadia/pull/318#discussion_r1726472814.

GeorgWa: Please let me know if I have some missunderstanding here.

Should the optimiztion order be a debug/experimental config or something the user really wants to change?

If it's encouraged to be configured by users, the order should be independent of the question of target/automatic. for example `order_of_optimization: [('rt','mobility')('ms2')('ms1)]` 

If this is something experimental for us so we can test different combinations, let's mark it like this.

alphadia/constants/default.yaml
GeorgWa: I assume this is experimental until we find something which always work?

I agree with Magnus, let's use different parameters, especially as they wont have the same metric update rules in the future. If we dont want to expose this in the gui, make it accessible, we don't ned to worry about having to many parameters.

alphadia/constants/default.yaml
odespard: My view is that it's something experimental. How should I mark it as such?

mschwoer: just add a comment that this is only for expert users?

alphadia/workflow/peptidecentric.py
mschwoer: I see .. should the "targeted optimizers always come before automatic optimizers." be added as a comment to the yaml?

mschwoer (approved):
LGTM

alphadia/workflow/peptidecentric.py
odespard: OK.

PR #319, opened 2024-08-22
Refactor parameter proposal

odespard:
Cf https://github.com/MannLabs/alphadia/pull/318#discussion_r1726450310.

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -86,10 +86,10 @@ class AutomaticOptimizer(BaseOptimizer):
         self.has_converged = False
         self.num_prev_optimizations = 0
         self.update_factor = workflow.config["optimization"][self.parameter_name][
-            "update_factor"
+            "automatic_update_factor"
         ]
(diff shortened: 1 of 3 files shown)


alphadia/workflow/optimization.py
```diff
@@ -226,9 +230,10 @@ def _propose_new_parameter(self, df):
         float
             The proposed new value for the search parameter.
 
-
         """
-        pass
+        return self.update_factor * self.workflow.calibration_manager.get_estimator(
+            self.estimator_group_name, self.estimator_name
+        ).ci(df, self.update_interval)
```
mschwoer, line 236: when reading `update_interval` I think of `time between updates` .. mb rename to `update_confidence_interval`?

mschwoer (approved):
LGTM

alphadia/workflow/optimization.py
odespard: I am wondering if we could find some better term than "confidence interval", because it isn't really a confidence interval in the normal sense. Perhaps ``update_percentile_range``?

PR #320, opened 2024-08-22
New RT convergence criteria

odespard:
Currently the graph of the proportion of precursors detected against RT error (which is used for automatic optimization) is relatively flat. This new set of convergence criteria and new way of finding the optimum (based on most narrow tolerance with proximity to the maximum). 

An example of the new behaviour is shown below, where the value taken forward for search is distinct from the value that maximises the number of precursors. 

image

NB: although this example shows the behaviour of the new rule, it is perhaps not the best illustration of why this change is useful. Often the curve is much flatter and the maximum is only slightly greater than all the other points, in which case choosing the narrower value is typically preferable. This change is also justified by the curve in our ground truth optimization (as shown in the precursor identifications in the plot below):
image

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -119,6 +119,12 @@ calibration:
   # TODO: remove this parameter
   norm_rt_mode: 'linear'
 
+  # the maximum number of fragments with correlation scores exceeding correlation_threshold to use for calibrating fragment mz (i.e. ms2)
+  max_fragments: 5000
+
+  # the correlation threshold for fragments used to calibrate fragment mz (i.e. ms2)
+  min_correlation: 0.7
+
 search_initial:
   # Number of peak groups identified in the convolution score to classify with target decoy comeptition
   initial_num_candidates: 1
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -190,7 +190,15 @@ search_output:
   file_format: "tsv"
 
 optimization:
+  # The order in which to perform optimization. Should be a list of lists of parameter names
+  # Example: [['ms1_error', 'ms2_error', 'rt_error', 'mobility_error']] means that all parameters are optimized simultaneously.
+  # Example: [["ms2_error"], ["rt_error"], ["ms1_error"], ["mobility_error"]] means that the parameters are optimized sequentially in the order given.
+  # Example: [["rt_error"], ["ms1_error", "ms2_error"]] means that first rt_error is optimized, then ms1_error and ms2_error are optimized simultaneously, and mobility_error is not optimized at all.
   order_of_optimization: null
+
+  # Parameters for the update rule for each parameter:
+  #   - update_interval: the percentile interval to use (as a decimal)
+  #   - update_factor: the factor by which to multiply the result from the percentile interval to get the new parameter value for the next round of search
   ms2_error:
       update_interval: 0.95
       update_factor: 1.0
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -192,17 +192,17 @@ search_output:
 optimization:
   order_of_optimization: null
   ms2_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   ms1_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   mobility_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
   rt_error:
-      update_interval: 0.99
-      update_factor: 1.1
+      update_interval: 0.95
+      update_factor: 1.0
 
 # configuration for the optimization manager
 # initial parameters, will be optimized
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -243,26 +243,40 @@ optimization:
   # Parameters for the update rule for each parameter:
   #   - update_percentile_range: the percentile interval to use (as a decimal)
   #   - update_factor: the factor by which to multiply the result from the percentile interval to get the new parameter value for the next round of search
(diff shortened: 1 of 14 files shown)


alphadia/constants/default.yaml
```diff
@@ -254,6 +254,8 @@ optimization:
       targeted_update_factor: 1.0
       automatic_update_interval: 0.99
       automatic_update_factor: 1.1
+      maximal_decrease: 0.8
```
mschwoer, line 257: could you document here what these mean ?

alphadia/workflow/optimization.py
```diff
@@ -260,6 +234,54 @@ def _check_convergence(self):
...
+        self.workflow.optimization_manager.fit(
+            {"classifier_version": classifier_version_at_optimum}
+        )
+        self.workflow.optimization_manager.fit(
+            {"score_cutoff": score_cutoff_at_optimum}
+        )
+        self.workflow.optimization_manager.fit({"fwhm_rt": fwhm_rt_at_optimum})
+        self.workflow.optimization_manager.fit(
+            {"fwhm_mobility": fwhm_mobility_at_optimum}
+        )
```
mschwoer, line 283: alternative ordering:

```python
optimal_parameter = self.history_df["parameter"].loc[index_of_optimum]
self.workflow.optimization_manager.fit({self.parameter_name: optimal_parameter})

classifier_version_at_optimum = ...
self.workflow.optimization_manager.fit(
    {"classifier_version": classifier_version_at_optimum}
)
...

```
unless, of course, the `.fit()` operations interfere with the `*_at_optimum` values

alphadia/workflow/optimization.py
```diff
@@ -400,13 +422,63 @@ def __init__(
...
+            "rt_error"
+        ]["minimum_proportion_of_maximum"]
         super().__init__(initial_parameter, workflow, reporter)
 
     def _get_feature_value(
         self, precursors_df: pd.DataFrame, fragments_df: pd.DataFrame
     ):
         return len(precursors_df) / self.workflow.optlock.total_elution_groups
 
+    def _check_convergence(self):
```
mschwoer, line 438: (nit) from a method that start with `check_` I would always assume that something is raised if the check fails.. 
`_has_converged()`?

alphadia/workflow/optimization.py
```diff
@@ -400,13 +422,63 @@ def __init__(
...
+        """
+        if len(self.history_df) <= 2:
+            return False
+
+        min_steps_reached = (
+            self.num_prev_optimizations
+            >= self.workflow.config["calibration"]["min_steps"]
+        )
+
+        feature_substantially_decreased = (
```
mschwoer, line 457: (nit) maybe `feature_history = self.history_df[self.feature_name]` improves readability even more?

alphadia/workflow/optimization.py
```diff
@@ -400,13 +422,63 @@ def __init__(
...
+            self.history_df[self.feature_name].iloc[-1]
+            < self.maximal_decrease * self.history_df[self.feature_name].iloc[-2]
+            and self.history_df[self.feature_name].iloc[-1]
+            < self.maximal_decrease * self.history_df[self.feature_name].iloc[-3]
```
mschwoer, line 461: if you go to `-3` here, should the check in l. 449 also be `<=3`?

mschwoer (approved):
LGTM!

alphadia/workflow/optimization.py
odespard: No, because that will only stop returning False when we have at least three rows in the dataframe, at which point the -3 index should not pose any problems.

alphadia/workflow/optimization.py
odespard: That's already the name of the attribute set by a conditional if self._check_convergence() is True. The easiest thing might be to make just_converged a property with the current definition of self._check_convergence(), and then we could dispense with the function. This is probably another PR though.

alphadia/workflow/optimization.py
odespard: But I've now made it < 3 instead of <= 2 for clarity.

alphadia/workflow/optimization.py
```diff
@@ -400,13 +422,63 @@ def __init__(
...
+            self.history_df["parameter"].iloc[-1]
+            / self.history_df["parameter"].iloc[-2]
+            > self.minimum_proportion_of_maximum
+        )
+
+        return min_steps_reached and (
+            feature_substantially_decreased or parameter_not_substantially_changed
+        )
+
+    def _find_index_of_optimum(self):
```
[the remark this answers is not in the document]
odespard: I'll add this in a chained PR where I also make this new convergence rule 1) configurable, and 2) available for all parameters.

alphadia/constants/default.yaml
odespard: I'll add this in a chained PR where I also make this new convergence rule 1) configurable, and 2) available for all parameters.

odespard (commented):
I won't merge this until we merge the next chained PR (https://github.com/MannLabs/alphadia/pull/322), as I want to be able to use the old convergence criteria through the config.

alphadia/workflow/optimization.py
```diff
@@ -260,6 +234,58 @@ def _check_convergence(self):
...
+
+        classifier_version_at_optimum = self.history_df["classifier_version"].loc[
+            index_of_optimum
+        ]
+        self.workflow.optimization_manager.fit(
+            {"classifier_version": classifier_version_at_optimum}
+        )
+
+        optimal_parameter = self.history_df["parameter"].loc[index_of_optimum]
+        self.workflow.optimization_manager.fit({self.parameter_name: optimal_parameter})
```
mschwoer, line 272: (nit^3) 
turning the order of the first two (also in docstring) would make it even more readable :-)
```
        optimal_parameter = self.history_df["parameter"].loc[index_of_optimum]
        self.workflow.optimization_manager.fit({self.parameter_name: optimal_parameter})

        classifier_version_at_optimum = self.history_df["classifier_version"].loc[
            index_of_optimum
        ]
        self.workflow.optimization_manager.fit(
            {"classifier_version": classifier_version_at_optimum}
        )
```

mschwoer (approved):
LGTM!

PR #323, opened 2024-08-23
Make fragment calibration configurable

odespard:
Currently, the version of automatic calibration that has performed the best is from before we fixed the bug that artificially reduced the number of fragments we picked relative to our intention; it always picked the 500 best fragments. It is possible that calibrating on this higher-quality set meaningfully improves performance and this can be tested by making this parameter configurable. 

NB: I have not made the min_fragments configurable because I don't think that the min_fragments code actually has any effect on the way the code currently runs (if there are fewer fragments than min_fragments, calibration still proceeds with the lower number). I may therefore remove it soon.

### DIFF
--- a/alphadia/workflow/peptidecentric.py
+++ b/alphadia/workflow/peptidecentric.py
@@ -549,7 +549,10 @@ class PeptideCentricWorkflow(base.WorkflowBase):
             by="correlation", ascending=False
         )
         # Determine the number of fragments to keep
-        min_fragments, max_fragments = 500, self.config["calibration"]["max_fragments"]
+        min_fragments, max_fragments = (
+            500,
+            self.config["calibration"]["max_fragments"],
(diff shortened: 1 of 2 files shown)


alphadia/workflow/peptidecentric.py
```diff
@@ -549,8 +549,8 @@ def filter_dfs(self, precursor_df, fragments_df):
             by="correlation", ascending=False
         )
         # Determine the number of fragments to keep
-        min_fragments, max_fragments = 500, 5000
-        min_correlation = 0.7
+        min_fragments, max_fragments = 500, self.config["calibration"]["max_fragments"]
```
mschwoer, line 552: mb add a TODO repeating what you mentioned in the description (that `min_fragments` could possibly be removed)

mschwoer (approved):

PR #324, opened 2024-08-26
Refactor search parameter optimization

odespard:


alphadia/workflow/peptidecentric.py
```diff
@@ -416,77 +416,35 @@ def search_parameter_optimization(self):
...
-                    f"=== Step {current_step}, extracting elution groups {self.optlock.start_idx} to {self.optlock.stop_idx} ===",
-                    verbosity="progress",
-                )
 
-                feature_df, fragment_df = self.extract_batch(
-                    self.optlock.batch_library.precursor_df,
-                    self.optlock.batch_library.fragment_df,
-                )
-                self.optlock.update_with_extraction(feature_df, fragment_df)
+                log_string(f"Starting optimization step {current_step}.")
```
mschwoer, line 420: l.402:
my IDE wants me to define `self.optlock` in the constructor, so I guess it also wants you to do the same ;-)

image

just initialize it with `None` there: it's a good practice to have all instance variables defined at one place

also, check the type annotations: 
image

mschwoer: now going down that rabbit hole: you might also check the return types of these, the seem off ;-)
image

alphadia/workflow/peptidecentric.py
```diff
@@ -416,77 +416,35 @@ def search_parameter_optimization(self):
...
-                self.optlock.update_with_fdr(precursor_df)
-
-                log_string(
-                    f"=== FDR correction performed with classifier version {self.optimization_manager.classifier_version} ===",
-                )
-
-                self.log_precursor_df(precursor_df)
+                precursor_df = self.process_batch()
 
                 if not self.optlock.has_target_num_precursors:
```
mschwoer, line 424: 👍 still a huge method, but much easier to get now

alphadia/workflow/peptidecentric.py
```diff
@@ -514,6 +472,68 @@ def search_parameter_optimization(self):
 
         self.save_managers()
 
+    def process_batch(self):
```
mschwoer, line 475: could all of these new methods be private? also, please add type hints and (concise) docstrings ..

mschwoer (approved):
LGTM!

PR #328, opened 2024-08-28
allow definition of target RT as fraction

odespard:


alphadia/workflow/peptidecentric.py
```diff
@@ -284,9 +284,15 @@ def get_ordered_optimizers(self):
             )
 
         if config_search["target_rt_tolerance"] > 0:
+            gradient_length = self.dia_data.rt_values.max()
+            target_rt_error = (
+                config_search["target_rt_tolerance"]
```
mschwoer, line 289: please add this contract as a comment to the default config and also to the UI

odespard: I have done so although I would appreciate your checking if what I have done with the .js file is remotely sensible.

mschwoer: I think this is quite the right place to add this info :-)

mschwoer (approved):

PR #329, opened 2024-08-28
Use batch index at optimum

odespard:
At present, if the optimization proceeds beyond the optimum, it may cause an increase in the batch index. The next round will then start from the higher batch instead of the batch index of the optimum. This code changes it to take the batch index of the optimum, at the cost of potentially getting the batch library twice if these are the same. (But the time impact of this is negligible and the benefits can be significant.)

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -314,7 +314,7 @@ class AutomaticOptimizer(BaseOptimizer):
             score cutoff,
             FWHM_RT,
             and FWHM_mobility
-        at the optimal parameter. Also updates the optlock with the smallest batch index.
+        at the optimal parameter. Also updates the optlock with the batch index at the optimum.
 
         """
         index_of_optimum = self._find_index_of_optimum()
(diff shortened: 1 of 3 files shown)


alphadia/workflow/optimization.py
```diff
@@ -780,6 +771,8 @@ def set_batch_dfs(self, eg_idxes):
         eg_idxes: np.ndarray
             The elution group indexes to use for the next round of optimization.
```
mschwoer, line 772: mb document what the default is

mschwoer (approved):

alphadia/workflow/optimization.py
```diff
@@ -343,6 +344,9 @@ def _update_optimization_manager(self):
             {"fwhm_mobility": fwhm_mobility_at_optimum}
         )
 
+        batch_index_at_optimum = self.history_df["batch_idx"].loc[index_of_optimum]
```
mschwoer, line 347: Maybe add this info from the description as a comment here
`take the batch index of the optimum, at the cost of potentially getting the batch library twice if these are the same. (But the time impact of this is negligible and the benefits can be significant.)...`

PR #330, opened 2024-08-29
Additional convergence criterion

odespard:
At present, optimizers need to have reached the target number of precursors before they can converge. This poses a problem if the optimizer tries a value so narrow that it is difficult to identify the target number. However, such values are not likely to optimize the feature value (indeed, if the feature value is well chosen they shouldn't, as we should not pick parameter values that reduce precursor counts so much). This PR introduces a new set of methods which record when optimization has been skipped and considers optimization to have converged if the optimizer is skipped too many times (as defined by max_skips in the calibration field of the config). The max_skips field has a default value of 1 and is unlikely to be worth changing as long as we use an exponential batch plan, since skipping an optimizer more than once implies that performance in precursors identifications is at least roughly twice as bad as before.

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -49,10 +49,10 @@ class BaseOptimizer(ABC):
         ----------
 
         precursors_df: pd.DataFrame
-            The precursor dataframe for the search
+            The filtered precursor dataframe for the search (see peptidecentric.PeptideCentricWorkflow.filter_dfs).
 
         fragments_df: pd.DataFrame
-            The fragment dataframe for the search
+            The filtered fragment dataframe for the search (see peptidecentric.PeptideCentricWorkflow.filter_dfs).
 
 
         """
@@ -88,6 +88,7 @@ class AutomaticOptimizer(BaseOptimizer):
         self.workflow.optimization_manager.fit({self.parameter_name: initial_parameter})
         self.has_converged = False
         self.num_prev_optimizations = 0
+        self._num_consecutive_skips = 0
         self.update_factor = workflow.config["optimization"][self.parameter_name][
             "automatic_update_factor"
         ]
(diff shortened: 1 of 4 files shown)


alphadia/workflow/optimization.py
```diff
@@ -49,16 +49,21 @@ def step(self, precursors_df: pd.DataFrame, fragments_df: pd.DataFrame):
...
         fragments_df: pd.DataFrame
-            The fragment dataframe for the search
+            The filtered fragment dataframe for the search (see peptidecentric.PeptideCentricWorkflow.filter_dfs).
 
 
         """
 
         pass
 
+    @abstractmethod
```
mschwoer, line 62: you could just make this an ordinary method, and change the docstring to `Default implementation: skipping of optimization is not recorded`
(and remove it from the subclass where it's implemented as `pass`)

alphadia/workflow/optimization.py
```diff
@@ -88,6 +93,7 @@ def __init__(
         self.workflow.optimization_manager.fit({self.parameter_name: initial_parameter})
         self.has_converged = False
         self.num_prev_optimizations = 0
+        self.num_consecutive_skips = 0
```
mschwoer, line 96: could this (and others?) be private?

alphadia/workflow/optimization.py
```diff
@@ -165,6 +157,17 @@ def step(
                 verbosity="progress",
             )
 
+    def skip(self):
+        """Increments the internal counter for the number of consecutive skips and checks if the optimization should be stopped."""
+        self.num_consecutive_skips += 1
+        if self._unnecessary_to_continue:
```
mschwoer, line 163: `_unnecessary_to_continue` is a bit sloppy, mb find a better name for this concept?

mschwoer (approved):

alphadia/workflow/optimization.py
odespard: I do refer to them outside the class, but only for logging purposes. Should I just move the logging inside the class?

alphadia/workflow/optimization.py
odespard: Good idea.

alphadia/workflow/optimization.py
odespard: Yes, I agree. I tried for a while and couldn't come up with anything, but I will try again.

Maybe ``_batch_substantially_bigger``?

mschwoer: mb also add some of the PR description here in order to fully get this

odespard: Or ``_large_drop_in_precursor_proportion``?

alphadia/workflow/optimization.py
odespard: I have done this.

PR #331, opened 2024-08-29
Refactor convergence rules

odespard:
Updates convergence rules to keep current functionality but allow them also work with negative values (useful in principle if we want to use a feature which reaches a minimum at the desired optimum -- we can just maximize the negative) and cope if parameter values increase during optimization (which is unusual but can happen). Also splits the behaviour currently activated by favour_narrower_parameter in two: continuing until a significant decrease is enabled now by try_narrower_values in the calibration field, while choosing the narrowest parameter within a certain deviation from the maximum is enabled by favour_narrower_optimum in the calibration field.

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -101,14 +101,20 @@ class AutomaticOptimizer(BaseOptimizer):
             self.parameter_name
         ]["automatic_update_percentile_range"]
 
-        self.favour_narrower_parameter = workflow.config["optimization"][
-            self.parameter_name
-        ]["favour_narrower_parameter"]
+        self.try_narrower_values = workflow.config["optimization"][self.parameter_name][
+            "try_narrower_values"
+        ]
 
-        if self.favour_narrower_parameter:
+        if self.try_narrower_values:
             self.maximal_decrease = workflow.config["optimization"][
                 self.parameter_name
             ]["maximal_decrease"]
+
+        self.favour_narrower_optimum = workflow.config["optimization"][
+            self.parameter_name
+        ]["favour_narrower_optimum"]
+
+        if self.favour_narrower_optimum:
             self.maximum_decrease_from_maximum = workflow.config["optimization"][
                 self.parameter_name
             ]["maximum_decrease_from_maximum"]
@@ -277,10 +283,10 @@ class AutomaticOptimizer(BaseOptimizer):
     @property
     def _just_converged(self):
         """Optimization should stop if continued narrowing of the parameter is not improving the feature value.
-        If self.favour_narrower_parameter is False:
+        If self.try_narrower_values is False:
             1) This function checks if the previous rounds of optimization have led to a meaningful improvement in the feature value.
             2) If so, it continues optimization and appends the proposed new parameter to the list of parameters. If not, it stops optimization and sets the optimal parameter attribute.
-        If self.favour_narrower_parameter is True:
+        If self.try_narrower_values is True:
             1) This function checks if the previous rounds of optimization have led to a meaningful disimprovement in the feature value or if the parameter has not changed substantially.
             2) If not, it continues optimization and appends the proposed new parameter to the list of parameters. If so, it stops optimization and sets the optimal parameter attribute.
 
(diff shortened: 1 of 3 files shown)


alphadia/workflow/optimization.py
```diff
@@ -101,17 +101,23 @@ def __init__(
             self.parameter_name
         ]["automatic_update_percentile_range"]
 
-        self.favour_narrower_parameter = workflow.config["optimization"][
-            self.parameter_name
-        ]["favour_narrower_parameter"]
+        self.try_narrower_values = workflow.config["optimization"][self.parameter_name][
```
mschwoer, line 104: could be private? (also others)

alphadia/workflow/optimization.py
```diff
@@ -101,17 +101,23 @@ def __init__(
...
                 self.parameter_name
             ]["maximal_decrease"]
-            self.minimum_proportion_of_maximum = workflow.config["optimization"][
+
+        self.favour_narrower_optimum = workflow.config["optimization"][
+            self.parameter_name
+        ]["favour_narrower_optimum"]
+
+        if self.favour_narrower_optimum:
+            self.maximum_decrease_from_maximum = workflow.config["optimization"][
```
mschwoer, line 118:
```
self.maximum_decrease_from_maximum = workflow.config["optimization"][
                self.parameter_name
            ]["maximum_decrease_from_maximum"] if self.favour_narrower_optimum else None
```
(otherwise, the class would have different attribute fields depending on config .. (check also above)

alphadia/workflow/optimization.py
```diff
@@ -296,23 +302,26 @@ def _just_converged(self):
...
 
             feature_history = self.history_df[self.feature_name]
             feature_substantially_decreased = (
-                feature_history.iloc[-1]
-                < self.maximal_decrease * feature_history.iloc[-2]
-                and feature_history.iloc[-1]
-                < self.maximal_decrease * feature_history.iloc[-3]
-            )
+                feature_history.iloc[-1] - feature_history.iloc[-2]
+            ) / np.abs(feature_history.iloc[-2]) < -self.maximal_decrease and (
```
mschwoer, line 314: so `self.maximal_decrease` carries a minus sign? could we get rid of that (direction is already encoded in the name, from which I would also guess that it is always the same sign?)

alphadia/workflow/optimization.py
```diff
@@ -296,23 +302,26 @@ def _just_converged(self):
...
+                feature_history.iloc[-1] - feature_history.iloc[-3]
+            ) / np.abs(feature_history.iloc[-3]) < -self.maximal_decrease
 
             parameter_history = self.history_df["parameter"]
             parameter_not_substantially_changed = (
-                parameter_history.iloc[-1] / parameter_history.iloc[-2] > 0.95
+                np.abs(
+                    (parameter_history.iloc[-1] - parameter_history.iloc[-2])
```
mschwoer, line 321: mb introduce
```
last_value = parameter_history.iloc[-1]
second_last_value = ..
third_last_value = ..
``` ?

mschwoer (approved):

alphadia/workflow/optimization.py
odespard: What do you mean by getting rid of it? I could force the use to use a negative sign in the config definition, but I thought it was easier to get the user to put in a positive value and then take the negative here. I am happy to change it if you think the former is more intuitive though.

mschwoer: ah, I did not read the full expression 
`    (feature_history.iloc[-1] - feature_history.iloc[-2]) / np.abs(feature_history.iloc[-2]) < -self.maximal_decrease`
having `maximal_decrease` positive is fine ..

mb turn it around:
   ` ( feature_history.iloc[-2] - feature_history.iloc[-1]) / np.abs(feature_history.iloc[-2]) > self.maximal_decrease`

odespard: I could do that. But I think it makes it harder to understand rather than easier (at least, I mentally reverse the operation in my head when trying to figure out what the code is doing).

PR #333, opened 2024-08-29
Proceed if insufficient precursors

odespard:
Proceeds to extraction with library values (or current optimum, although this is unlikely to happen) if there are insufficient precursors.


mschwoer (approved):

GeorgWa (approved):
LGTM

PR #334, opened 2024-08-30
updated documentation

odespard:
Documentation for optimization and calibration rewritten to reflect current procedure more accurately.


docs/methods/search_parameter_optimization.md
```diff
@@ -0,0 +1,131 @@
...
+
+<img src="../_static/images/methods_optimization.png" width="100%" height="auto">
+
+Optimization can be performed either in either a targeted or automatic manner. In targeted optimization, the search space is progressively narrowed until a target tolerance is reached for a given parameter. In automatic optimization, the search space is progressively narrowed until an internal algorithm detects that further narrowing will reduce performance, at which point the optimal value is selected for search. Automatic optimization can be triggered by setting the target tolerance to 0.0 in the GUI, or by setting it to any nonpositive value in the command line. It is possible to use targeted optimization for some parameters and automatic optimization for others; currently, it is recommended to use targeted optimization for precursor m/z, fragment m/z and ion mobility, and automatic optimization for retention time.
```
mschwoer, line 9: "performance" -> what is meant here? speed? quality?

mschwoer, line 9: > setting the target tolerance to 0.0 in the GUI, or by setting it to any nonpositive value in the command line.
is the behaviour not unified between GUI & CLI?

docs/methods/search_parameter_optimization.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Optimization
+For optimizing the search space, tolerances like retention time, ion mobility and m/z ratios need to be reduced. The goal is to cover the expected spectrum space but reduce it as much as possible to accelerate search and gain statsitical power. Search starts with initial tolerances as defined in `search_initial`.  For targeted optimization, the 95% deviation after calibration is adopted as the new tolerance until the target tolerances defined in the `search` section are reached. For automatic optimization, the 99% deviation plus 10% of the absolute value of the tolerance is adopted as the new tolerance, and search continues until parameter-specific convergence rules are met.
+
+The optimization is finished as soon as the minimum number of steps `min_steps` has passed and all tolerances have either 1) reached the target tolerances defined in `search` if using targeted optimization, or 2) have converged if using automatic optimization.
+
+## Configuring calibration and optimization
+The configuration below will perform targeted optimization of precursor m/z, fragment m/z and ion mobility, and automatic optimization of retention time.
+
+```yaml
```
mschwoer, line 28: Showing the config here is a good service to the reader, but it is likely that this will get outdated .. 
you may choose to add all these comments to the actual default config, and just point to the `calibration`, `search_initial`, and `search` keys here..
but it's also fine to leave it like this
(I had this discussion with @GeorgWa already ;-))

docs/methods/search_parameter_optimization.md
```diff
@@ -0,0 +1,131 @@
...
+  target_rt_tolerance: 0
+
+  quant_window: 3
+  quant_all: True
+```
+
+## Calibration using LOESS
+Individual properties like the retention time deviate from their library values and need to be calibrated (a). As a nonlinear but stable method, locally estimated scatterplot smoothing (LOESS) using both density and uniformly distributed kernels is used. (b) A collection of polynomial kernels is fitted to uniformly distributed subregions of the data. These consist of first and second degree polynomials basis functions of the calibratable property. (c) The individual functions are combined and smoothed using tricubic weights. (d) Combining the kernels with their weighting functions allows to approximate the systematic deviation of the data locally. (e), The sum of the weighted kernels can then be used for continuous approximation and calibration of retention times. The architecture is built on the scikit-learn package and can be configured to use different hyperparameters and arbitrary predictors for calibration.
+
+<img src="../_static/images/methods_loess.png" width="100%" height="auto">
```
mschwoer, line 76: figure missing?

docs/methods/search_parameter_optimization.md
```diff
@@ -0,0 +1,131 @@
+# Search Parameter Optimization
+
+## Calibration and optimization
+### Overall process
+The first step of every alphaDIA run is the optimization of search parameters and the calibration of the empirical or fully predicted spectral library to the observed values. This step has two main purposes: 1) removing the systematic deviation of observed and library values, and 2) optimizing the size of the search space to reflect the expected deviation from the library values. For DIA search this means calibration and optimization of certain parameters: retention time, ion mobility, precursor m/z and fragment m/z. The process of iterative calibration and optimization is illustrated below.
```
mschwoer, line 5: (nit) `AlphaDIA` .. also, there's a typo in `statsitical` few paragraphs below ;-)

mschwoer (approved):
Really cool documentation!

docs/methods/search_parameter_optimization.md
odespard: Perhaps this is an unnecessary level of detail for the documentation. The behaviour is unified in that it all goes through the config in the same way. However, in the GUI one is restricted to setting the tolerance to non-negative values, whereas one is not so restricted when one specifies the config. However, I think mentioning this in the documentation is more confusing than helpful so I will remove it.

docs/methods/search_parameter_optimization.md
odespard: I can see a figure here.

docs/methods/search_parameter_optimization.md
odespard: I've simplified what I'm showing so as to include only relevant sections of each field in the config. I'm not sure that this fully addresses the issue you raise, but I think it makes it slightly better than before.

docs/methods/search_parameter_optimization.md
odespard: I've rephrased it.

PR #337, opened 2024-09-09
Fix incorrect mobility outputaccumulator

mo-sameh:
I’ve fixed the issue where the default columns for mobility, RT, and m/z were incorrect due to differing indexes when copying the columns. I’ve also added a test case to test for the expected behavior.


GeorgWa (approved):
LGTM

PR #340, opened 2024-09-10
Update GUI for release

GeorgWa:
- Align default GUI and yaml parameters
- fix peptdeep checkpoint bug

### DIFF
--- a/alphadia/libtransform.py
+++ b/alphadia/libtransform.py
@@ -328,7 +328,7 @@ class PeptDeepPrediction(ProcessingStep):
             logging.info(f"Loading PeptDeep models of type {self.peptdeep_model_type}")
             model_mgr.load_installed_models(self.peptdeep_model_type)
 
-        if self.peptdeep_model_path is not None:
+        if self.peptdeep_model_path is not None and self.peptdeep_model_path != "":
             if not os.path.exists(self.peptdeep_model_path):
                 raise ValueError(
                     f"PeptDeep model checkpoint folder {self.peptdeep_model_path} does not exist"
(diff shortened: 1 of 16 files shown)


alphadia/libtransform.py
```diff
@@ -328,7 +328,7 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
             logging.info(f"Loading PeptDeep models of type {self.peptdeep_model_type}")
             model_mgr.load_installed_models(self.peptdeep_model_type)
 
-        if self.peptdeep_model_path is not None:
+        if self.peptdeep_model_path is not None and self.peptdeep_model_path != "":
```
mschwoer, line 331: `if self.peptdeep_model_path`?

mschwoer (approved):
LGTM

alphadia/libtransform.py
mschwoer: should we run e2e tests here?
(classic 'submarine' PR: disguised as UI change, but changes default parameters ;-))

GeorgWa: Yes, to some degreee :D But all of these parameters are redefined in the e2e tests. So we only change to enable Oscars optimization by default.

alphadia/libtransform.py
```diff
@@ -324,11 +324,11 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
         model_mgr = ModelManager(device=device)
 
         # will load other model than default generic
-        if self.peptdeep_model_type is not None:
+        if self.peptdeep_model_type:
             logging.info(f"Loading PeptDeep models of type {self.peptdeep_model_type}")
             model_mgr.load_installed_models(self.peptdeep_model_type)
 
-        if self.peptdeep_model_path is not None and self.peptdeep_model_path != "":
+        if self.peptdeep_model_path and self.peptdeep_model_path != "":
```
mschwoer, line 331: just `if self.peptdeep_model_path` will do .. `""` is falsy

PR #341, opened 2024-09-10
BUG whitespace in config path

GeorgWa:


mschwoer (approved):
Alternative solution: show random results to people who use whitespaces in paths.
edit: whitespaces at the end of file names are fine, ofc

PR #348, opened 2024-09-14
BUG optimization lock, fix #347

### LINKED ISSUE #347: Optimization fails with: zero-size array to reduction operation minimum which has no identity
**Describe the bug**
It looks like optimization can fail when the number of precursors get's very low due to FDR instability. In this example, optimization was started as lots of precursors where found. during optimization, only 199 precursors where found.

```python
1:19:14.677819 INFO: Traceback (most recent call last):
  File "alphadia\cli.py", line 342, in run
    plan.run()
  File "alphadia\planning.py", line 365, in run
    raise e
  File "alphadia\planning.py", line 338, in run
    workflow.search_parameter_optimization()
  File "alphadia\workflow\peptidecentric.py", line 453, in search_parameter_optimization
    self.optlock.update()
  File "alphadia\workflow\optimization.py", line 878, in update
    self.decrease_batch_idx()
  File "alphadia\workflow\optimization.py", line 863, in decrease_batch_idx
    smallest_value = np.min(
                     ^^^^^^^
  File "numpy\core\fromnumeric.py", line 2953, in min
  File "numpy\core\fromnumeric.py", line 88, in _wrapreduction
ValueError: zero-size array to reduction operation minimum which has no identity

1:19:14.678819 ERROR: zero-size array to reduction operation minimum which has no identity
```

**Expected behavior**
aborting optimization with the best solution so far

**Version (please complete the following information):**
alphaDIA 1.8.0

GeorgWa:
- light refactoring
- fix bug where the optimization lock should be optimized

### DIFF
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -818,7 +818,7 @@ class OptimizationLock:
         self.batch_plan = plan
 
     def batches_remaining(self):
-        return self.batch_idx + 1 > len(self.batch_plan)
+        return self.batch_idx + 1 < len(self.batch_plan)
 
     def update_with_extraction(
         self, feature_df: pd.DataFrame, fragment_df: pd.DataFrame
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -848,7 +848,7 @@ class OptimizationLock:
             The precursor dataframe for the current batch of the optimization lock (from workflow.perform_fdr).
         """
 
-        self._precursor_at_fdr_count = np.sum(precursor_df["qval"] <= 0.01)
+        self._precursor_at_fdr_count = np.sum(precursor_df["qval"] < 0.01)
         self.has_target_num_precursors = (
             self._precursor_at_fdr_count >= self._precursor_target_count
         )
--- a/alphadia/workflow/optimization.py
+++ b/alphadia/workflow/optimization.py
@@ -848,7 +848,9 @@ class OptimizationLock:
             The precursor dataframe for the current batch of the optimization lock (from workflow.perform_fdr).
         """
 
-        self._precursor_at_fdr_count = np.sum(precursor_df["qval"] < 0.01)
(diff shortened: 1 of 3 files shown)


alphadia/workflow/optimization.py
```diff
@@ -822,9 +848,10 @@ def update_with_fdr(self, precursor_df: pd.DataFrame):
             The precursor dataframe for the current batch of the optimization lock (from workflow.perform_fdr).
         """
 
-        self.count = len(precursor_df[precursor_df["qval"] < 0.01])
-
-        self.has_target_num_precursors = self.count >= self.target_count
+        self._precursor_at_fdr_count = np.sum(precursor_df["qval"] < 0.01)
```
odespard, line 851: Should the precursor_df be filtered here to include only targets and only one channel, e.g. ``np.sum((precursor_df["qval"] < 0.01) & (precursor_df["decoy"] == 0) & (precursor_df["channel"] == channel)
)``?
Alternatively it could perhaps be filtered in the peptidecentric.py file.

odespard (approved):
This looks sensible.

alphadia/workflow/optimization.py
GeorgWa: added the filter

alphadia/workflow/optimization.py
```diff
@@ -768,6 +770,27 @@ def __init__(self, library: SpecLibFlat, config: dict):
...
+    @property
+    def fragments_df(self) -> pd.DataFrame:
+        return pd.concat(self.fragment_dfs)
+
+    @property
+    def start_idx(self) -> int:
+        if self.has_target_num_precursors:
+            return 0
+        elif self.batch_idx >= len(self.batch_plan):
+            raise NoOptimizationLockTargetError()  # This should never be triggered since introduction of the BaseOptimizer.proceed_with_insufficient_precursors method and associated code, and could be removed.
```
mschwoer, line 786: so.. remove it?

mschwoer (approved):

PR #349, opened 2024-09-16
Allow correct processing of empty files

GeorgWa:
Makes sure empty files are correctly processed without failure.
- missing `_update_history()` function
- make `_find_index_of_optimum` compatible with failed optimization
- add testcase to test failed optimization with `AutomaticOptimizer`
- handle lfq and mbr failure gracefully

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -262,6 +262,10 @@ class QuantBuilder:
 
         _intensity_df.sort_values(by=group_column, inplace=True, ignore_index=True)
 
+        _intensity_df.to_csv(
+            "/Users/georgwallmann/Downloads/intensity_df.tsv", sep="\t", index=False
+        )
+
         lfq_df = lfqutils.index_and_log_transform_input_df(_intensity_df)
         lfq_df = lfqutils.remove_allnan_rows_input_df(lfq_df)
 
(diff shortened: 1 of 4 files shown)


alphadia/outputtransform.py
```diff
@@ -813,6 +813,16 @@ def build_lfq_tables(
                 group_column=group,
             )
 
+            # remove all rows for testing
+            group_intensity_df = group_intensity_df.iloc[0:0]
```
mschwoer, line 817: accidental commit?

alphadia/outputtransform.py
```diff
@@ -834,11 +844,11 @@ def build_lfq_tables(
                     file_format=self.config["search_output"]["file_format"],
                 )
 
-        protein_df_melted = lfq_df.melt(
-            id_vars="pg", var_name="run", value_name="intensity"
-        )
-
-        psm_df = psm_df.merge(protein_df_melted, on=["pg", "run"], how="left")
+        if len(lfq_df) > 0 and len(lfq_df.columns) > 2:
```
mschwoer, line 847: we are reusing `lfq_df` from above, right? could this be made more robust, or explicit?
(cf. also the comment "# IMPORTANT: 'pg' has to be the last group in the list as this will be reused")

GeorgWa: Excellent task for claude

mschwoer (approved):
LGTM

PR #358, opened 2024-10-20
BUG zero division on zero intensity detector events

GeorgWa:
Fixes a zero division error on zero intensity detector events in TimsTOF data

### DIFF
--- a/alphadia/data/bruker.py
+++ b/alphadia/data/bruker.py
@@ -567,7 +567,7 @@ class TimsTOFTransposeJIT:
         n_precursor_indices = len(unique_precursor_index)
         n_tof_slices = len(tof_limits)
 
-        # scan valuesa
+        # scan values
         mobility_start = int(scan_limits[0, 0])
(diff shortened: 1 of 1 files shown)


alphadia/data/bruker.py
```diff
@@ -586,6 +586,16 @@ def assemble_push(
             dtype=np.float32,
         )
 
+        # intensities below HIGH_EPSILON will be set to zero
```
mschwoer, line 589: there's a similar code in `AlphaRawJIT` .. maybe introduce two little helpers methods
`lower_zero_cutoff` and `zero_safe_division` ?

note there's also an alternative idiom https://stackoverflow.com/a/68118106

GeorgWa: In every other part of the code I would agree. But this is quite performance critical, so I want to have an explicit compile time constant.

mschwoer: I doubt that importing this as a method would give any performance overhead here :-)

PR #366, opened 2024-11-06
autofixes for linting

mschwoer:
going towards stronger liniting: apply auto-fixes for extended ruleset ("ALL" minus exceptions)

Exceptions are taken from the alphakraken

### DIFF
--- a/alphadia/calibration/property.py
+++ b/alphadia/calibration/property.py
@@ -172,7 +172,7 @@ class Calibration:
             self.function.fit(input_values, target_value)
             self.is_fitted = True
         except Exception as e:
-            logging.error(f"Could not fit estimator {self.name}: {e}")
+            logging.exception(f"Could not fit estimator {self.name}: {e}")
             return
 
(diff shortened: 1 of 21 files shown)


alphadia/calibration/property.py
```diff
@@ -172,7 +172,7 @@ def fit(self, dataframe: pd.DataFrame, plot: bool = False, **kwargs):
             self.function.fit(input_values, target_value)
             self.is_fitted = True
         except Exception as e:
-            logging.error(f"Could not fit estimator {self.name}: {e}")
+            logging.exception(f"Could not fit estimator {self.name}: {e}")
```
GeorgWa, line 175: This will take some time until it reaches muscle memory :D

GeorgWa (approved):
LGTM

alphadia/calibration/property.py
mschwoer: doesn't need to: it's an autofix :-D

PR #368, opened 2024-11-08
add gradient_length_m to output

mschwoer:
Adds gradient length in minutes to the `stats.tsv`

image

@GeorgWa To me this feels a bit like a hack, as I add this information to every entry in the psm table, although it's a property of the whole run. Happy to take suggestions.

update: added an alternative implementation sketch (marked with `# alternative`, changes to `SearchPlanOutput` not done but should be obvious) .. if we will have more output data on the "run" level, this might be more sustainable

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -982,6 +982,8 @@ def _build_run_stat_df(
             base_dict["rt_error"] = np.nan
             base_dict["mobility_error"] = np.nan
 
+        base_dict["gradient_length_m"] = channel_df["gradient_length"].max() / 60
+
         out_df.append(base_dict)
 
     return pd.DataFrame(out_df)
--- a/alphadia/planning.py
+++ b/alphadia/planning.py
@@ -312,6 +312,7 @@ class Plan:
 
         workflow_folder_list = []
 
+        run_info = {}  # alternative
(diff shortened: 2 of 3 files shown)


alphadia/outputtransform.py
```diff
@@ -982,6 +982,8 @@ def _build_run_stat_df(
             base_dict["rt_error"] = np.nan
             base_dict["mobility_error"] = np.nan
 
+        base_dict["gradient_length_m"] = channel_df["gradient_length"].max() / 60
```
mschwoer, line 985: alternatively: `rt_max` .. and we should mb keep the unit in seconds (I guess the other rt observables rest is seconds?)

GeorgWa:
Yes, I agree. 

This is the role of the `Manager` objects are being used for right now. We could either attach this to the optimisation manager or move this into a separate manager. I’m already logging a couple of raw file stats on loading and we could convert this into a dict.

In the other PR #369 I’ve started to use semantic naming like `optimization.ms2_error`. We could add all of them to the stats table like `raw.rt_max`.

alphadia/outputtransform.py
jalew188: I prefer using minutes, for example, we say 21 min gradient, 11 min gradient, but not 21*60 seconds gradient.

PR #371, opened 2024-11-13
Rename progress folder

mschwoer:
- rename the `.progress` folder to `quant`
- rename the input parameter `custom-quant-dir` to `quant-dir`

### DIFF
--- a/alphadia/cli.py
+++ b/alphadia/cli.py
@@ -96,7 +96,7 @@ parser.add_argument(
     default="{}",
 )
 parser.add_argument(
-    "--custom-quant-dir",
+    "--quant-dir",
     type=str,
     help="Directory to save the quantification results (psm & frag parquet files) to be reused in a distributed search",
     nargs="?",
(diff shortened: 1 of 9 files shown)


alphadia/cli.py
```diff
@@ -380,7 +382,7 @@ def run(*args, **kwargs):
             library_path=library_path,
             fasta_path_list=fasta_path_list,
             config=config,
-            custom_quant_dir=custom_quant_dir,
+            quant_path=quant_dir,
```
GeorgWa, line 385: can't we call both of them quant_dir for simplicity?

GeorgWa (approved):
LGTM

alphadia/cli.py
mschwoer: "_path" is more correct and should be used throughout the code base (cf. also the `# TODO rename all output_directory, output_folder => output_path, quant_dir->quant_path (except cli parameter)`)

imho it's fine to name the cli parameter `quant-dir` as this might be more familiar to users .. very early on this should be called `quant_path` then

GeorgWa: Makes sense to me

PR #372, opened 2024-11-13
introduce rawfilemanager

mschwoer:
Introduce rawfilemanager to be able to persist statistics on raw files


alphadia/data/stats.py
```diff
@@ -1,45 +0,0 @@
-import logging
```
mschwoer, line 1: this went to RawFileManager.calc_stats & _log_stats

alphadia/workflow/base.py
```diff
@@ -82,9 +85,18 @@ def load(
         self.reporter.context.__enter__()
         self.reporter.log_event("section_start", {"name": "Initialize Workflow"})
 
-        self.reporter.log_event("loading_data", {"progress": 0})
         # load the raw data
-        self._dia_data = self._get_dia_data_object(dia_data_path)
+        self.reporter.log_event("loading_data", {"progress": 0})
+        raw_file_manager = RawFileManager(
```
mschwoer, line 90: this is a very short-lived manager, is this alright?

alphadia/data/alpharaw_wrapper.py
```diff
@@ -4,23 +4,13 @@
...
-from alpharaw import mzml as alpharawmzml
-
-# TODO fix: "import resolves to its containing file"
-from alpharaw import sciex as alpharawsciex
-
-# alpha family imports
-from alpharaw import (
-    thermo as alpharawthermo,
-)
+from alpharaw import mzml as alpharaw_mzml
```
mschwoer, line 9: kind of unrelated, but this totally confused my IDE

GeorgWa: mission accomplished :D

alphadia/workflow/base.py
GeorgWa: Yes

alphadia/workflow/base.py
```diff
@@ -82,9 +83,18 @@ def load(
...
+            self.config,
+            path=os.path.join(self.path, self.RAW_FILE_MANAGER_PATH),
+            reporter=self.reporter,
+        )
+
+        self._dia_data = raw_file_manager.get_dia_data_object(dia_data_path)
+        raw_file_manager.calc_stats(self._dia_data)
```
GeorgWa, line 95: What do you think about always saving the stats as part of `raw_file_manager.get_dia_data_object` ?

alphadia/workflow/managers/raw_file_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        -------
+        typing.Union[bruker.TimsTOFTranspose, thermo.Thermo],
+            TimsTOFTranspose object containing the DIA data
+
+        """
+        file_extension = os.path.splitext(dia_data_path)[1]
+
+        is_wsl = self._config["general"]["wsl"]
+        if is_wsl:
+            # copy file to /tmp # TODO why is that?
```
GeorgWa, line 53: There was some weird behjaviour in WSL.
I think I will actually drop WSL support soon, as it's not needed anymore.

alphadia/workflow/managers/raw_file_manager.py
```diff
@@ -0,0 +1,151 @@
...
+class RawFileManager(BaseManager):
+    def __init__(
+        self,
+        config: None | Config = None,
+        path: None | str = None,
+        load_from_file: bool = False,
+        **kwargs,
+    ):
+        """Contains and updates timing information for the portions of the workflow."""
+        self.stats = {}  # needs to be before super().__init__ to avoid overwriting when loading
```
GeorgWa, line 24: Is there a solution for this? sounds like a potential pitfall?

GeorgWa (approved):
Looks good! We need to test this with e2e for sure.

alphadia/workflow/managers/raw_file_manager.py
mschwoer: indeed I fell into that pit yesterday (LLMs that hallucinate that pickling a dictionary with mixed data types is problematic didn't really help ;-)) ..
option # 1

`if not hasattr(self, '_config'):
    self._config: Config = config`
ugly, uses hasattr

option # 2:
`if not load_from_file: 
    self.stats = {}`
at least this makes it more explicit than just the comment but does not prevent new users falling into the pit..

option # 3: 
each class has an explicit list of attributes to pickle, and some mechanism to instantiate those only in case its not pickled .. overkill imho..


I would go for option 2 and add a comment to the super class as well

alphadia/workflow/managers/raw_file_manager.py
mschwoer: I am happy to drop it right away?

GeorgWa: Please do! But it’s across the GUI, CLI etc. so I would do it in a separate PR

alphadia/workflow/managers/raw_file_manager.py
GeorgWa: Sounds good 🤔 Ideally I would like to move away from pickling at some point, which could include this. Its just not an important feature.

PR #373, opened 2024-11-13
Add rt length to output ii

mschwoer:
First guess at what should be added in terms of raw file stats.

Also some renamings and refactoring of the pickle code.


nbs/tutorial_nbs/logging.ipynb
```diff
@@ -21,11 +21,11 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "0:00:00.000142 \u001b[32;20mPROGRESS: test\u001b[0m\n",
+      "0:00:00.000142 \u001B[32;20mPROGRESS: test\u001B[0m\n",
```
mschwoer, line 24: not sure what this was?

alphadia/outputtransform.py
```diff
@@ -1024,6 +1025,30 @@ def _build_run_stat_df(
...
+                path=raw_file_manager_path, load_from_file=True
+            )
+            prefix = "raw."
+            stats = raw_file_manager.stats
+
+            # deliberately mapping explicitly to avoid coupling stats to the output too tightly
+            base_dict[f"{prefix}rt_limit_min"] = stats["rt_limit_min"]
+            base_dict[f"{prefix}rt_limit_max"] = stats["rt_limit_max"]
+
+            base_dict[f"{prefix}rt_duration_sec"] = stats["rt_duration_sec"]
```
GeorgWa, line 1043: I actually agree with Feng that we should convert the values to minutes. We can store them as minutes from the beginning.

alphadia/workflow/manager.py
```diff
@@ -82,51 +82,51 @@ def is_fitted(self, value):
...
-            try:
-                with open(self.path, "wb") as f:
-                    pickle.dump(self, f)
-            except Exception as e:
-                self.reporter.log_string(
-                    f"Failed to save {self.__class__.__name__} to {self.path}: {str(e)}",
-                    verbosity="error",
-                )
-                # Log the full traceback
+        if self.path is None:
```
GeorgWa, line 85: Thanks for the refactor 😄

alphadia/workflow/managers/raw_file_manager.py
```diff
@@ -22,13 +22,15 @@ def __init__(
...
         """Contains and updates timing information for the portions of the workflow."""
         self.stats = {}  # needs to be before super().__init__ to avoid overwriting when loading
-
         super().__init__(path=path, load_from_file=load_from_file, **kwargs)
-        self.reporter.log_string(f"Initializing {self.__class__.__name__}")
-        self.reporter.log_event("initializing", {"name": f"{self.__class__.__name__}"})
 
         self._config: Config = config
 
+        # deliberately not saving the actual raw data here to avoid the saved manager file being too large
```
GeorgWa, line 29: This is just a comment why the dia_data goes back into `PeptideCentric` and doesn't stay here as one might expect?

nbs/tutorial_nbs/logging.ipynb
GeorgWa, line 24: We have colored logging. I don't know why but it looks like your machine likes a different ANSI escape than mine???

GeorgWa (approved):
I think switching to time units as minutes is an important one. Everything else is good to go.

alphadia/workflow/managers/raw_file_manager.py
mschwoer: yes it is

GeorgWa (approved):
LGTM

PR #390, opened 2024-11-27
Misc improvements

mschwoer:
some minor changes (consider them as suggestions)

see commit messages for details

### DIFF
--- a/alphadia/planning.py
+++ b/alphadia/planning.py
@@ -49,16 +49,23 @@ class Plan:
 
         Parameters
         ----------
-        raw_data : list
+
+        output_folder : str
+            output folder to save the results
+
+        raw_path_list : list
             list of input file locations
 
-        config_path : str, optional
+        library_path : str, optional
+            path to the spectral library file. If not provided, the library is built from fasta files
+
+        fasta_path_list : list, optional
+            list of fasta file locations to build the library from
+
+        config_base_path : str, optional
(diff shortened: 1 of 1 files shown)


alphadia/planning.py
```diff
@@ -158,14 +157,22 @@ def spectral_library(self, spectral_library: SpecLibFlat) -> None:
...
         logger.progress(f"{'alphatims':<15} : {alphatims.__version__:}")
         logger.progress(f"{'alpharaw':<15} : {alpharaw.__version__}")
         logger.progress(f"{'alphabase':<15} : {alphabase.__version__}")
         logger.progress(f"{'alphapeptdeep':<15} : {peptdeep.__version__}")
         logger.progress(f"{'directlfq':<15} : {directlfq.__version__}")
         logger.progress("===================================================")
 
+        logger.progress("================= Pip Environment =================")
+        pip_env = [
+            f"{dist.metadata['Name']}=={dist.version}"
```
GeorgWa, line 170: I thought about using debug Level logging. Bit I guess it doesnt hurt to have this by default.

mschwoer: you will never need this until you need it, so I would leave it on by default

what about the `progress` level anyway? could this not be INFO?

GeorgWa: The idea is that progress has the important stuff and is colored green so you can find it faster when checking output

PR #391, opened 2024-11-27
Misc fixes 2

mschwoer:
- unify import of njit
- fix typo in config (breaking ;-))

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -53,7 +53,7 @@ library_prediction:
 
 # define custom alphabase modifications not part of unimod or alphabase
 # also used for decoy channels
-custom_modififcations:
+custom_modifications:
   # Dimethyl @K channel decoy
   Dimethyl:d12@K:
     composition: H(-2)2H(8)13C(2)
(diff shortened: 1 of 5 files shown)


alphadia/constants/default.yaml
```diff
@@ -53,7 +53,7 @@ library_prediction:
 
 # define custom alphabase modifications not part of unimod or alphabase
 # also used for decoy channels
-custom_modififcations:
```
GeorgWa, line 56: It’s breaking in theory :D alphabase can’t use the custom mods with multiprocessing yet. So it’s not being used (yet)

alphadia/quadrupole.py
```diff
@@ -13,7 +6,7 @@
 from alphadia import utils
 
 
-@alphatims.utils.njit
```
GeorgWa, line 16: Did you check the alphatims function? I wonder why I did this in the first place ( I already see some weird numbs thing happening :D )

GeorgWa (approved):
LGTM

alphadia/quadrupole.py
mschwoer: this is the `@alphatims.utils.njit` decorator
```
def njit(_func=None, *args, **kwargs):
    import numba
    return numba.njit(_func, *args, **kwargs)
```

GeorgWa: So much value

PR #402, opened 2024-12-13
Refactor multistep search

mschwoer:
- move logging of logo etc out of `Plan`
- move logging out of cli
- introduce a string constants class
- some minor refactorings

### DIFF
--- a/alphadia/constants/keys.py
+++ b/alphadia/constants/keys.py
@@ -1,0 +1,24 @@
+class ConstantsClass(type):
+    """A metaclass for classes that should only contain string constants."""
+
+    def __setattr__(self, name, value):
+        raise TypeError("Constants class cannot be modified")
+
+    def get_values(cls):
+        """Get all user-defined string values of the class."""
+        return [
+            value
+            for key, value in cls.__dict__.items()
+            if not key.startswith("__") and isinstance(value, str)
+        ]
+
+
+class OutputKeys(metaclass=ConstantsClass):
+    """String constants for reading and writing output columns."""
+
+    # optimization
+    OPTIMIZATION_PREFIX = "optimization."
+    MS1_ERROR = "ms1_error"
+    MS2_ERROR = "ms2_error"
+    RT_ERROR = "rt_error"
+    MOBILITY_ERROR = "mobility_error"
--- a/alphadia/constants/keys.py
+++ b/alphadia/constants/keys.py
@@ -13,8 +13,8 @@ class ConstantsClass(type):
         ]
 
 
-class OutputKeys(metaclass=ConstantsClass):
-    """String constants for reading and writing output columns."""
+class StatOutputKeys(metaclass=ConstantsClass):
+    """String constants for reading and writing output columns for the `stat` file."""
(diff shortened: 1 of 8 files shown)


alphadia/outputtransform.py
```diff
@@ -981,15 +977,22 @@ def _build_run_stat_df(
             optimization_manager = manager.OptimizationManager(
                 path=optimization_manager_path
             )
-            optimization_stats[MS2_ERROR] = optimization_manager.ms2_error
-            optimization_stats[MS1_ERROR] = optimization_manager.ms1_error
-            optimization_stats["rt_error"] = optimization_manager.rt_error
-            optimization_stats["mobility_error"] = optimization_manager.mobility_error
+            optimization_stats[OutputKeys.MS2_ERROR] = optimization_manager.ms2_error
```
mschwoer, line 980: @GeorgWa I'd suggest using this idiom for all string constants short- to mid-term

alphadia/constants/keys.py
```diff
@@ -0,0 +1,24 @@
...
+    def get_values(cls):
+        """Get all user-defined string values of the class."""
+        return [
+            value
+            for key, value in cls.__dict__.items()
+            if not key.startswith("__") and isinstance(value, str)
+        ]
+
+
+class OutputKeys(metaclass=ConstantsClass):
```
GeorgWa, line 16: Should we rename this so it reflects it's used in the stats table?
like `StatsOutputKeys`

alphadia/workflow/manager.py
```diff
@@ -494,12 +494,16 @@ def __init__(
                 "fwhm_mobility": config["optimization_manager"]["fwhm_mobility"],
                 "score_cutoff": config["optimization_manager"]["score_cutoff"],
             }
-            self.__dict__.update(initial_parameters)
+            self.__dict__.update(
```
GeorgWa, line 497: instead of using the magic method? ok

alphadia/workflow/manager.py
```diff
@@ -494,12 +494,16 @@ def __init__(
...
+                initial_parameters
+            )  # TODO either store this as a dict or in individual instance variables
 
             for key, value in initial_parameters.items():
                 self.reporter.log_string(f"initial parameter: {key} = {value}")
 
-    def fit(self, update_dict):
+    def fit(
+        self, update_dict
+    ):  # TODO siblings' implementations have different signatures
```
GeorgWa, line 506: yes.. this is unfortunately necessary :/ Should we name the functions differently in this case?

GeorgWa (approved):
Looks good to me

alphadia/workflow/manager.py
mschwoer: any preferences? `store this as a dict` or `individual instance variables`?

alphadia/workflow/manager.py
mschwoer: this would be a solution, yes .. otherwise using inheritance does not make too much sense (but maybe it does not in the first place?)

PR #405, opened 2024-12-13
update e2e settings

GeorgWa:
e2e test settings for multistep didn't include the necessary fixed mod.


mschwoer:
Found this out already here, by adapting the e2e settings to the dimethyl tutorial: https://github.com/MannLabs/alphadia/pull/403/files#diff-bcf417d0d8a7a2763216e4a493c6dd6d04ed5f441b474b954a962bcdab1a1fe5

which version should we keep for the e2e?

GeorgWa:
We can keep your settings. 👍🏻 There might be small performance differences but this isn't the goal of the e2e test yest

mschwoer:
decision: using version in https://github.com/MannLabs/alphadia/pull/405

Issue #406, opened 2024-12-14
MS2 transfer learning broken


GeorgWa:
# Description

Performing transfer learning using the alphadia `FinetuneManager` fails for MS2 with zero PCC already on the forward pass.
I think this might be related to peptdeep expecting a different library order.

```
2024-12-14 09:47:46>  Model tested on validation dataset with the following metrics:
2024-12-14 09:47:46>  l1_loss                       : 0.0353
2024-12-14 09:47:46>  PCC-mean                      : 0.0000
2024-12-14 09:47:46>  COS-mean                      : 0.0000
2024-12-14 09:47:46>  SA-mean                       : 0.0000
2024-12-14 09:47:46>  SPC-mean                      : 0.2157
2024-12-14 09:47:46>  Fine-tuning MS2 model with the following settings:
2024-12-14 09:47:46>  Train fraction:      0.70     Train size:      65216     
2024-12-14 09:47:46>  Validation fraction: 0.20     Validation size: 18633     
2024-12-14 09:47:46>  Test fraction:       0.10     Test size:       9317      
2024-12-14 09:48:25>  Epoch 0   Lr: 0.00020   Training loss: 0.0243   validation loss: 0.0353
2024-12-14 09:48:59>  Epoch 3   Lr: 0.00050   Training loss: 0.0219   validation loss: 0.0353
2024-12-14 09:49:38>  Epoch 6   Lr: 0.00060   Training loss: 0.0227   validation loss: 0.0353
2024-12-14 09:50:17>  Epoch 9   Lr: 0.00030   Training loss: 0.0226   validation loss: 0.0353
```

# Steps to reproduce

I've uploaded some validated test data here:
https://datashare.biochem.mpg.de/s/4mdf5PMHsjLXGZe

Using the finetuning notebook and the current dev version of the alphax packages should reproduce the behaviour.

PR #407, opened 2024-12-15
Fix ms2 transferlearning ms2 bug

mo-sameh:
A fix for #406 

This might be a good place to explain what was going on:
- From my experiences, in peptdeep when using the prediction functionality directly from the model class (instead of using predict_all in the model manager), it expects the precursors to be sorted by nAA. If they aren’t sorted, you end up with an all-zero fragments dataframe. That’s why I added the df_refinement step in this PR to handle this issue.

- When the precursors are sorted by nAA, PeptDeep drops the old fragment indices and reinitializes them (in-place) based on the refined precursors. This caused problems when trying to compare predictions with a ground truth. This part of the issue was already handled but I wanted to explain why we are doing multiple preprocessing steps before prediction, and this commit specifically fixes the nAA sorting-related problem. 

This should address both issues and ensure things work as expected.

### DIFF
--- a/alphadia/transferlearning/train.py
+++ b/alphadia/transferlearning/train.py
@@ -5,6 +5,7 @@ import pandas as pd
 import torch
 from alphabase.peptide.fragment import remove_unused_fragments
 from alphabase.peptide.mobility import ccs_to_mobility_for_df, mobility_to_ccs_for_df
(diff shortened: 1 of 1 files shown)


GeorgWa:
Hi Mo, we got rid of development branch. Can you reopen the PR based of main?

PR #409, opened 2024-12-15
fix: zero predictions for ms2

mo-sameh:
A fix for https://github.com/MannLabs/alphadia/issues/406

This might be a good place to explain what was going on:

From my experiences, in peptdeep when using the prediction functionality directly from the model class (instead of using predict_all in the model manager), it expects the precursors to be sorted by nAA. If they aren’t sorted, you end up with an all-zero fragments dataframe. That’s why I added the df_refinement step in this PR to handle this issue.

When the precursors are sorted by nAA, PeptDeep drops the old fragment indices and reinitializes them (in-place) based on the refined precursors. This caused problems when trying to compare predictions with a ground truth. This part of the issue was already handled but I wanted to explain why we are doing multiple preprocessing steps before prediction, and this commit specifically fixes the nAA sorting-related problem.

This should address both issues and ensure things work as expected.

### DIFF
--- a/alphadia/transferlearning/train.py
+++ b/alphadia/transferlearning/train.py
@@ -5,6 +5,7 @@ import pandas as pd
 import torch
 from alphabase.peptide.fragment import remove_unused_fragments
 from alphabase.peptide.mobility import ccs_to_mobility_for_df, mobility_to_ccs_for_df
+from alphabase.peptide.precursor import refine_precursor_df
 from peptdeep.model.charge import ChargeModelForModAASeq
 from peptdeep.model.model_interface import CallbackHandler, LR_SchedulerInterface
 from peptdeep.pretrained_models import ModelManager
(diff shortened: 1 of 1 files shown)


alphadia/transferlearning/train.py
```diff
@@ -564,13 +565,14 @@ def finetune_ms2(
         test_intensity_df = test_intensity_df[0]
 
         # Prepare order for peptdeep prediction
-
+        val_psm_df = refine_precursor_df(val_psm_df)
```
mschwoer, line 568: could you add brief version of the PR comment here (explaining why this is required)?

alphadia/transferlearning/train.py
```diff
@@ -5,6 +5,7 @@
 import torch
```
mschwoer, line 5: should we also take action in peptdeep? e.g. log a warning if the precursors are not sorted?

mschwoer (approved):
LGTM!

PR #410, opened 2024-12-16
Change config handling

mschwoer:
- change the handling of the config. The idea is to have a config dumped to disk that is fully self-contained, i.e. no additional CLI parameters are required.
- strictly forbid to add new keys (as they would not be understood by AlphaDIA anyway), this requires change the `modifications` section
- refactor what is passed to SearchPlan

See the updated docs for the motivation :-)

PR #411, opened 2024-12-19
Refactor config

mschwoer:
Refactor the config class:
- separate merging and printing (up to the `tracking_config` variable)
- add more fine grained unit tests


alphadia/workflow/config.py
```diff
@@ -563,30 +82,219 @@ def update(self, experiments: list["Config"], print_modifications: bool = True):
...
+
+    Raises:
+        ValueError in these cases:
+        - a key is not found in the target_config
+        - the type of the update value does not match the type of the target value
+        - an item is not found in the target_config
+    """
+    for key, update_value in update_config.items():
+        if key not in target_config:
+            raise ValueError(f"Key not found in target_config: '{key}'")
```
GeorgWa, line 163: Should we make this more explicit? 
Something like: Updating configs does not allow defining new keys.

GeorgWa (approved):
Much cleaner implementation! LGTM

alphadia/workflow/config.py
mschwoer: done here https://github.com/MannLabs/alphadia/pull/410/files#diff-089981037ebf31cbb2b167a4089dc8541ff70653e39ccaa71e148cf871362a95R81-R94

PR #414, opened 2024-12-20
add more unit tests for config

mschwoer:
Add more unit tests for config

### DIFF
--- a/tests/unit_tests/test_config.py
+++ b/tests/unit_tests/test_config.py
@@ -159,3 +159,138 @@ def test_get_modifications_table():
     target = pd.read_csv(StringIO(target_tsv), sep="\t", index_col=0)
 
     pd.testing.assert_frame_equal(table, target)
+
+
+generic_default_config = """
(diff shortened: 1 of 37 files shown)


tests/unit_tests/test_config.py
```diff
@@ -159,3 +159,138 @@ def test_get_modifications_table():
...
+    config_1.update([config_2, config_3], print_modifications=True)
+
+    assert config_1.to_dict() == expected_generic_default_config_dict | {
+        "simple_value_int": 2,
+        "simple_value_float": 5.0,
+        "simple_value_str": "six",
+    }
+
+
+def test_config_update_advanced():
```
GeorgWa, line 252: How I understand it we test for more involved update logic:
- lists of different lengths are matched to same lengths
- lists of dicts are updated if the names match
I guess this is changed in later PR's because we decided to stay with simpler logic

GeorgWa (approved):
LGTM

PR #415, opened 2024-12-20
Refactor config2

mschwoer:
- make the config an actual dict, simplify the way it is initialized
- remove ability to set config_base_path


GeorgWa (approved):
LGTM

PR #418, opened 2024-12-20
fix fdr bug

mschwoer:
"fix" the FDR "bug" by reducing number of threads for FDR classification task
apparently it does not scale well:

large task (linux machine w/ 32 physical cores, 64 is hyperthreading):
image

small task:
image


alphadia/fdr.py
```diff
@@ -113,8 +114,18 @@ def perform_fdr(
         X, y, test_size=0.2
     )
 
+    # the classifier does not scale above 2 threads also for large problems
+    num_threads = torch.get_num_threads()
```
GeorgWa, line 118: should we also allow passing num_threads?

GeorgWa (approved):
LGTM

alphadia/fdr.py
GeorgWa: No, just that perform_fdr accepts a threads parameter

GeorgWa: For example if we want to perform benchmarking in the Future

GeorgWa (approved):
Let’s merge this

PR #419, opened 2024-12-20
fix the nans in accumulated transferlib bug

mo-sameh:
Changing from loc to iloc to fix a bug where the first row of fragments for all precursors (except the first) contained only NaNs. This issue occurred because, when using loc with a slice, both endpoints are included by default (see: [Pandas documentation on advanced indexing](https://pandas.pydata.org/docs/user_guide/advanced.html#advanced-endpoints-are-inclusive)).

In contrast, by definition of frag_start and frag_stop indices specifies that frag_stop is exclusive, which aligns with the behavior of iloc.

### DIFF
--- a/.github/workflows/_run_tests.yml
+++ b/.github/workflows/_run_tests.yml
@@ -34,6 +34,11 @@ jobs:
     - name: Conda info
       shell: bash -l {0}
       run: conda info
+    - name: Install mono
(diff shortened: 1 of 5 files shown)


GeorgWa (approved):
Could you add a testcase for the outputaccumulator in a separate PR?
I will merge this briefly and check the release :)

PR #420, opened 2024-12-20
no mono for windows

mschwoer:


.github/workflows/_run_tests.yml
```diff
@@ -35,17 +35,17 @@ jobs:
...
       shell: bash -l {0}
       run: |
         conda install mono
-
     - name: Perform pip installation with all stable dependencies
       shell: bash -l {0}
       run: |
         cd misc
-        . ./${{ inputs.install-script }} alphadia ${{ inputs.python-version }}
+        . ./${{ inputs.install-script }} alphadia ${{ inputs.python-version }} ${{ runner.os != 'windows-latest' }}
```
GeorgWa, line 46: Could this fail if it’s windows-not-so-late?

GeorgWa (approved):
LGTM

.github/workflows/_run_tests.yml
mschwoer: made it future proof (assuming 'future' will go full circle one day)

PR #421, opened 2024-12-31
Transfer-learning allow additional fragtypes

mo-sameh:
Using the functionality introduced in peptdeep [PR](https://github.com/MannLabs/alphapeptdeep/pull/226). 
To allow finetuning a pretrained model to be able to predict different number of charged_frag_types while using the pretrained backbone as a starting point for the weights.

### DIFF
--- a/alphadia/transferlearning/train.py
+++ b/alphadia/transferlearning/train.py
@@ -3,10 +3,9 @@ import logging
 import numpy as np
 import pandas as pd
 import torch
-from alphabase.peptide.fragment import remove_unused_fragments
+from alphabase.peptide.fragment import get_charged_frag_types, remove_unused_fragments
 from alphabase.peptide.mobility import ccs_to_mobility_for_df, mobility_to_ccs_for_df
 from alphabase.peptide.precursor import refine_precursor_df
-from alphabase.peptide.fragment import get_charged_frag_types
 from peptdeep.model.charge import ChargeModelForModAASeq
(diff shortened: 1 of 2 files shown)


alphadia/transferlearning/train.py
```diff
@@ -233,11 +234,16 @@ def __init__(
...
+        self.charged_frag_types = (
+            get_charged_frag_types(fragment_types, max_charge)
+            if fragment_types
+            else None
+        )
         assert (
             self._train_fraction + self._validation_fraction + self._test_fraction
             <= 1.0
         ), "The sum of the train, validation and test fractions should be less than or equal to 1.0"
+        super().__init__(mask_modloss, device, self.charged_frag_types)
```
mschwoer, line 246: I would prefer calling this init still at the top..
```
def ...(..):
    charged_frag_types = (
            get_charged_frag_types(fragment_types, max_charge)
            if fragment_types
            else None
        )
   super().__init__(mask_modloss, device, charged_frag_types)

   self.charged_frag_types = charged_frag_types
...
```

I did not find a proper resource, but I feel the parent init should be called "as early as possible"
https://stackoverflow.com/a/77975817

nbs/tutorial_nbs/additional_frags.ipynb
```diff
@@ -0,0 +1,480 @@
+{
```
mschwoer, line 1: A small initial description of what the purpose of this nb is would be nice

mschwoer (approved):
LGTM

PR #431, opened 2025-01-16
Add two step classifier

anna-charlotte:
Adding a two-step-classifier, that's to be used with a logistic regression followed by a neural network classifier, as inspired by DIA-NN. With this we aim to increase sensitivity, particularly in samples with only few peptides present, such as single cell samples.

Steps:
- [x] add a `TwoStepClassifier`
- [x] add a parameter to config to use the two-step-classifier
- [x] adjust call in `FDRManager.fit_predict()`, where the classifier is trained
- [x] add tests


alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
         """
 
 
+class TwoStepClassifier(Classifier):
+    def __init__(
+        self,
+        first_classifier: Classifier,
+        second_classifier: Classifier,
+        train_on_top_n: int = 1,
+        first_fdr: float = 0.6,
```
GeorgWa, line 236: maybe `first_fdr_cutoff` or `first_fdr_filter`

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+        return df_targets
+
+    @classmethod
+    def _update_classifier(
+        cls, classifier, df_, x_cols, y_col, fdr, group_columns
+    ) -> None:
+        X = df_[x_cols]
+        y = df_[y_col]
+        df = df_.copy()
+        if hasattr(classifier, "fitted") and classifier.fitted:
```
GeorgWa, line 321: why do we need the hasattr? could we solve this through the Classifier base class?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+
+        return df_targets
+
+    @classmethod
+    def _update_classifier(
+        cls, classifier, df_, x_cols, y_col, fdr, group_columns
+    ) -> None:
+        X = df_[x_cols]
+        y = df_[y_col]
+        df = df_.copy()
```
GeorgWa, line 320: I know it's annoying :D Do we need this copy?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+            df_subset["proba"] = self.second_classifier.predict_proba(x_subset)[:, 1]
+
+        df_subset = get_entries_below_fdr(
+            df_subset, self.second_fdr, group_columns
+        )  # , remove_decoys=True)
+        print(
+            f"After q-val extraction, after NN (second_fdr={self.second_fdr}): {df_subset.shape}"
+        )
+
+        df_subset_2 = get_target_decoy_partners(df_subset, df_)
```
GeorgWa, line 300: can we find more clear variable names for subset and subset_2?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+            previous_n_precursors = -1
+
+        classifier.fit(X, y)
+        classifier._fitted = True
+
+        df["proba"] = classifier.predict_proba(df[x_cols].to_numpy())[:, 1]
+        df_targets = get_entries_below_fdr(df, fdr, group_columns)
+
+        current_n_precursors = len(df_targets)
+        if previous_n_precursors > current_n_precursors:
```
GeorgWa, line 336: At the moment we have to run the first classifier twice. Could we just store the value from the first run and reuse it here?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
         """
 
 
+class TwoStepClassifier(Classifier):
```
GeorgWa, line 230: I think we don't need to implement the `Classifier` base class here.

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+                "Reverting linear classifier to the previous version, as the new one performed worse."
+            )
+            classifier.from_state_dict(saved_state_dict)
+
+    @property
+    def fitted(self) -> bool:
+        """Return whether both classifiers have been fitted."""
+        return self.second_classifier.fitted
+
+    def fit(self, x: np.ndarray, y: np.ndarray) -> None:
```
GeorgWa, line 347: I think we can skip this for now if we drop the `Classifier` parent class

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+        # # Filter data based on first FDR threshold
+        # mask = probs >= (1 - self.first_fdr)
+        # x_filtered = x[mask]
+        # y_filtered = y[mask]
+
+        # # Fit second classifier on filtered data
+        # self.second_classifier.fit(x_filtered, y_filtered)
+        pass
+
+    def predict(self, x: np.ndarray) -> np.ndarray:
```
GeorgWa, line 372: this too?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+            Input data
+
+        Returns
+        -------
+        np.ndarray
+            Predicted class labels
+        """
+        return np.argmax(self.predict_proba(x), axis=1)
+
+    def predict_proba(self, x: np.ndarray) -> np.ndarray:
```
GeorgWa, line 387: this too?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+
+def get_target_decoy_partners(df_sub, df_all):
+    group_by = ["rank", "elution_group_idx"]
+    valid_tuples = df_sub[group_by]
+    matching_rows = df_all.merge(valid_tuples, on=group_by, how="inner")
+
+    return matching_rows
+
+
+def keep_best(
```
GeorgWa, line 452: was this function copied from the old implementation?

alphadia/fdr.py
```diff
@@ -168,248 +168,18 @@ def perform_fdr(
...
-        The q-values of the PSMs.
-    """
-
-    y_test_proba = classifier.predict_proba(X_test)[:, 1]
-
-    y_train_proba = classifier.predict_proba(X_train)[:, 1]
-
-    fpr_test, tpr_test, thresholds_test = sklearn.metrics.roc_curve(
-        y_test, y_test_proba
+    df.dropna(subset=available_columns, inplace=True)
```
mschwoer, line 178: could we avoid the inplace operation?

alphadia/fdr_utils.py
```diff
@@ -0,0 +1,260 @@
...
+import numpy as np
+
+# third party imports
+import pandas as pd
+import sklearn
+
+logger = logging.getLogger()
+
+
+def keep_best(
```
mschwoer, line 16: +1 for reorganizing code, but this makes it hard to spot any changes :-)
would it be a large effort to move this back to `fdr.py` for now and do the reordering (=just moving) later (or: before) in a dedicated PR?

alphadia/fdrexperimental.py
```diff
@@ -7,13 +7,133 @@
...
+    target_values = 1 - _df[decoy_column].values
+    decoy_cumsum = np.cumsum(_df[decoy_column].values)
+    target_cumsum = np.cumsum(target_values)
+    fdr_values = decoy_cumsum / target_cumsum
+    _df[qval_column] = fdr_to_q_values(fdr_values)
+    return _df
+
+
+def apply_absolute_transformations(df: pd.DataFrame) -> pd.DataFrame:
+    df_transformed = df.copy()
```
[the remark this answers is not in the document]
mschwoer: couldn't we just copy the columns that are changed (assuming they are much smaller than the whole df) and merge them to the original df?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +125,347 @@ def from_state_dict(self, state_dict: dict):
...
+            y_col,
+            self.first_fdr,
+            group_columns,
+        )
+
+        return df_targets
+
+    @classmethod
+    def _update_classifier(
+        cls, classifier, df_, x_cols, y_col, fdr, group_columns
```
mschwoer, line 214: docstrings missing ;-) (also at other places)
also, I would avoid underscore prefixes in variable names unless they are really required (private variables, small-scoped clashes with built-ins, named ignores..)

mschwoer: @GeorgWa correct me if I'm wrong, but `df = df.copy()` would be okay in such a case?

alphadia/fdrexperimental.py
```diff
@@ -107,6 +125,347 @@ def from_state_dict(self, state_dict: dict):
         """
 
 
+class TwoStepClassifier(Classifier):
```
mschwoer, line 128: could all this new code be moved to a new module?

alphadia/fdr.py
mschwoer: 😄  (after reading @GeorgWa comments going in thte exact opposite direction)

alphadia/fdr_utils.py
GeorgWa: To my understanding the functions were copied as they are from https://github.com/MannLabs/alphadia/blob/da99596957d4025377d602e32be89dfb679eafbf/alphadia/fdr.py#L171

So there should be no changes here?

alphadia/fdrexperimental.py
GeorgWa: great point! lets roll out a `fdr` module.

Do you think we should do this in a separate PR @mschwoer to prevent overloading of this PR with moving of 'old' functionality?

mschwoer: new code I would put already into the new module .. moving old code would be a separate PR (cf https://github.com/MannLabs/alphadia/pull/431#discussion_r1918996623)

alphadia/fdrexperimental.py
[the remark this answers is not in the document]
GeorgWa: Yes, that would be an option. My favourite would be to move the abs() to the location where the feature is calculated in the first place. But this is a bit complicated for prototyping.

alphadia/fdrexperimental.py
GeorgWa: In this implementation we need the copy, good point!
I think we should only create a copy to store the indices for calculating the FDR etc. and not duplicate the features though.

alphadia/fdrexperimental.py
```diff
@@ -7,13 +7,133 @@
...
 import torch
 from sklearn import model_selection
+from sklearn.linear_model import LogisticRegression
+from sklearn.preprocessing import StandardScaler
 from torch import nn, optim
 from torchmetrics.classification import BinaryAUROC
 from tqdm import tqdm
 
 
+def fdr_to_q_values(fdr_values: np.ndarray):
```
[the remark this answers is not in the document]
anna-charlotte: first it was copied, then moved to `fdr_utils.py` due to circular import (which has now been resolved), so now everything back to pre-PR state ;)

alphadia/fdr.py
```diff
@@ -168,6 +170,23 @@ def perform_fdr(
     return psm_df
 
 
+def perform_fdr_new(
```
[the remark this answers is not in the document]
anna-charlotte: agree, done

alphadia/fdrexperimental.py
anna-charlotte: > was this function copied from here?

adressed in the previous comment :)

alphadia/fdr_utils.py
anna-charlotte: yes exactly, it was just moved over, due to a circular import issue. That one has been resolved now, so I moved it back to `alphadia/fdr.py` for now:)

alphadia/fdrexperimental.py
anna-charlotte: yeah, let me check all the `.copy()`'s and remove the redundant ones;)

alphadia/fdrexperimental.py
```diff
@@ -107,6 +227,383 @@ def from_state_dict(self, state_dict: dict):
...
+                :, 1
+            ]
+            df_subset = get_entries_below_fdr(
+                df, self.first_fdr, group_columns, remove_decoys=False
+            )
+            print(
+                f"After q-val extraction, after LinClassifier (first_fdr={self.first_fdr}): {df_subset.shape}"
+            )
+
+            self.second_classifier.batch_size = 500
```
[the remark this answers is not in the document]
anna-charlotte: yeah agree, this shouldn't stay here. It can't be moved to the initialization though, because we want to decrease the batch size only after the first_classifier has been fitted, due to the decreased size od the preselected df. 

I think this will be adressed with your [PR #435](https://github.com/MannLabs/alphadia/pull/435), shall I wait for that one to be merged and adjust here accordingly?

alphadia/workflow/manager.py
```diff
@@ -721,7 +766,8 @@ def load_classifier_store(self, path: None | str = None):
 
                 if classifier_hash not in self.classifier_store:
                     classifier = deepcopy(self.classifier_base)
-                    classifier.from_state_dict(torch.load(os.path.join(path, file)))
+                    with contextlib.suppress(Exception):
+                        classifier.from_state_dict(torch.load(os.path.join(path, file)))
```
anna-charlotte, line 770: @GeorgWa What is this `alphadia/constants/classifier/fa9945ae23db872d.pth` file that we are loading here, some pretrained model? Shall I store a similar file for the two-step-classifier?

GeorgWa: Yes exactly 👍🏻 We will do the same with the two step classifier eventually

alphadia/fdrexperimental.py
anna-charlotte: I moved the `TwoStepClassifier` and `LogisticRegressionModel` to a new module, which I called `fdr_analysis` for now, as it was name-clashing with the `fdr.py` file otherwise.

GeorgWa: could you move it to `fdrx`?

anna-charlotte:
@GeorgWa Should I add an e2e or performance test for the two step classifier, or just the unit tests?

alphadia/fdrx/models/logistic_regression.py
```diff
@@ -0,0 +1,128 @@
...
+        self._fitted = True
+
+    def predict(self, x: np.ndarray) -> np.ndarray:
+        """Predict the class of the data.
+
+        Parameters
+        ----------
+
+        x : np.array, dtype=float
+            Data of shape (n_samples, n_features).
```
mschwoer, line 47: at other places you are using capital `X` .. please choose one ;-) (check also what the rest of the code uses)

(and adapt also `x_scaled`)

alphadia/fdrx/models/logistic_regression.py
```diff
@@ -0,0 +1,128 @@
...
+        Parameters
+        ----------
+
+        dict : dict
+            Dictionary containing the state of the classifier.
+
+        """
+        self._fitted = state_dict["_fitted"]
+
+        if self.fitted:
```
mschwoer, line 118: please check if this should rather be `if self._fitted:`? if not, add a comment which deconfuses me :-)

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -0,0 +1,346 @@
...
+            best_precursor_count = current_target_count
+            best_result = predictions
+
+            # Update first classifier if enough confident predictions
+            if current_target_count > self.min_precursors_for_update:
+                self.update_first_classifier(
+                    df_filtered, df, x_cols, y_col, group_columns
+                )
+            else:
+                break
```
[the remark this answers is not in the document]
mschwoer: and log if we reached `max_iterations`? (could use the `for ... else` pattern here)

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -0,0 +1,346 @@
...
+            The minimum number of precursors required to update the first classifier.
+
+        """
+        self.first_classifier = first_classifier
+        self.second_classifier = second_classifier
+        self.first_fdr_cutoff = first_fdr_cutoff
+        self.second_fdr_cutoff = second_fdr_cutoff
+
+        self.min_precursors_for_update = min_precursors_for_update
+        self.train_on_top_n = train_on_top_n
```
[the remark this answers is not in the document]
anna-charlotte: Do you mean `self.min_precursors_for_update` and `self.train_on_top_n`? I think they could be private, I went with not private, following the pattern `BinaryClassifierLegacyNewBatching` where only `_fitted`is private. But i'm ok with either, shall i change it?

alphadia/fdrx/models/logistic_regression.py
anna-charlotte: I figured, since we implement `Classifier`, which has the [`@property def fitted(self): ...`](https://github.com/MannLabs/alphadia/blob/5f890028225361b98f3fb6263b31074b667f219b/alphadia/fdrexperimental.py#L31-L34), one would use `self._fitted` for setting, but `self.fitted` for accessing, is that right? If so, I'll add a comment, or otherwise change it 😃

alphadia/workflow/manager.py
```diff
@@ -776,7 +821,8 @@ def load_classifier_store(self, path: None | str = None):
 
                 if classifier_hash not in self.classifier_store:
                     classifier = deepcopy(self.classifier_base)
-                    classifier.from_state_dict(torch.load(os.path.join(path, file)))
+                    with contextlib.suppress(Exception):
```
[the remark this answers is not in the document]
anna-charlotte: I removed the suppresion and added a check for `TwoStepClassifier`-instance instead for now, and will add a pretrained model file in the future, such that the check can be removed again

alphadia/fdrx/models/logistic_regression.py
mschwoer: as it's the class accesing an instance variable, it's fine to use `self._fitted` (this is equivalent in terms of logic as here the property is a 1:1 wrapper around `self._fitted` to make it public)
more than "fine" actually: would prefer it for consistency :-)

@GeorgWa why do we have these properties anyway? they don't seem to be used

alphadia/fdrx/models/two_step_classifier.py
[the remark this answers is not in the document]
mschwoer: given it's a different class, I would favor correctness over consistency here.. so I'd prefer changing it

alphadia/fdrx/models/logistic_regression.py
anna-charlotte: yeah I figured we don't have the setter here, as we would want the `fitted` attribute to be read-only access for the user, otherwise would there be an advantage in this property being private, versus just haveing a `.fitted`attribute?

PR #434, opened 2025-01-16
Hotfix for alphaBase bug

GeorgWa:
This was previous undefined behaviour in alphaBase which now leads to a bug

### DIFF
--- a/alphadia/libtransform.py
+++ b/alphadia/libtransform.py
@@ -131,6 +131,10 @@ class DynamicLoader(ProcessingStep):
         else:
             raise ValueError(f"File type {file_type} not supported")
 
+        # TODO: this is a hack to get the charged_frag_types from the fragment_mz_df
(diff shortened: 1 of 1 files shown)


alphadia/libtransform.py
```diff
@@ -131,6 +131,10 @@ def forward(self, input_path: str) -> SpecLibBase:
         else:
             raise ValueError(f"File type {file_type} not supported")
 
+        # TODO: this is a hack to get the charged_frag_types from the fragment_mz_df
+        # this should be fixed ASAP in alphabase
```
mschwoer, line 135: please add a issue in AlphaBase and add the link to the code

mschwoer (approved):

alphadia/libtransform.py
GeorgWa: done

PR #437, opened 2025-01-17
Fix output path

### LINKED ISSUE #433: SearchPlan requires output dir in two places
**Describe the bug**
GIven the default debug notebook:

Alltough SearchPlan accepts an ouput folder, it does not set the value in the config.
Therefore:

```python
config = {
    "raw_paths": raw_data_path_list,
    "library_path": library_path,
    # "output_directory": output_folder,
}

plan = SearchPlan(output_folder, config=config)
```
fails with an inconclusive log message.

**To Reproduce**
Current dev version 

**Expected behavior**
- Error mesage which says that the mandatory key 'output directory' has not been set
- output dir is set even if the config doesnt contain the key OR SearchPlan doesn't allow to pass output dir

mschwoer:
fix #433

### DIFF
--- a/alphadia/workflow/config.py
+++ b/alphadia/workflow/config.py
@@ -14,6 +14,7 @@ from copy import deepcopy
 
 import yaml
 
+from alphadia.constants.keys import ConfigKeys
 from alphadia.exceptions import KeyAddedConfigError, TypeMismatchConfigError
 
 logger = logging.getLogger()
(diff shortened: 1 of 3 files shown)


alphadia/workflow/config.py
```diff
@@ -51,7 +52,9 @@ def to_json(self, path: str) -> None:
             json.dump(self.data, f)
 
     def __setitem__(self, key, item):
-        raise NotImplementedError("Use update() to update the config.")
+        if key != ConfigKeys.OUTPUT_DIRECTORY:
```
GeorgWa, line 55: This feels a bit hacky. What would be a potential general solution for this?

GeorgWa (approved):
LGTM

PR #445, opened 2025-01-20
fix bug in multistep config output dir

mschwoer:
fix bug in multistep search: output directory not propagated correctly

### DIFF
--- a/alphadia/search_step.py
+++ b/alphadia/search_step.py
@@ -62,6 +62,8 @@ class SearchStep:
         self._config = self._init_config(
             config, cli_config, extra_config, output_folder
         )
+        self._config.to_yaml(os.path.join(output_folder, "frozen_config.yaml"))
+
         logger.setLevel(logging.getLevelName(self._config["general"]["log_level"]))
 
         self.raw_path_list = self._config[ConfigKeys.RAW_PATHS]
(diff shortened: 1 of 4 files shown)


alphadia/search_step.py
```diff
@@ -117,16 +110,26 @@ def _init_config(
...
             # need to overwrite user-defined output folder here to have correct value in config dump
-            extra_config[ConfigKeys.OUTPUT_DIRECTORY] = output_folder
+            extra_config_update[ConfigKeys.OUTPUT_DIRECTORY] = output_folder
             config_updates.append(extra_config_update)
 
-        config.update(config_updates, do_print=True)
+        if config_updates:
+            config.update(config_updates, do_print=True)
 
         if config.get(ConfigKeys.OUTPUT_DIRECTORY, None) is None:
```
GeorgWa, line 119: In what case are we missing the output folder?
If it wasn't set in any config and only provided as `arg` to SearchPlan?

GeorgWa (approved):
LGTM

alphadia/search_step.py
mschwoer: exactly .. in order to have it in the `frozen_config.yaml` and also in the construction of the `self._quant_path`, we set it here explicitly

PR #446, opened 2025-01-21
add stricter linting

mschwoer:
add a second layer of linting, that acts on selected files .. the goal is to reduce the blacklist bit by bit ;-)


GeorgWa (approved):
LGTM

anna-charlotte (approved):
lgtm

PR #448, opened 2025-01-23
Fix fasta in cli

mschwoer:
#prfromhell

### DIFF
--- a/tests/unit_tests/test_cli.py
+++ b/tests/unit_tests/test_cli.py
@@ -1,15 +1,75 @@
 """This module provides unit tests for alphadia.cli."""
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, mock_open, patch
 
-from alphadia.cli import run
+import yaml
+
+from alphadia.cli import _get_config_from_args, _get_from_args_or_config, run
+
+# TODO add tests for _get_raw_path_list_from_args_and_config
+
+
+def test_get_config_from_args():
+    """Test the _get_config_from_args function correctly merges configs."""
+    mock_args = MagicMock(config="config.yaml", config_dict='{"key3": "value3"}')
+
+    yaml_content = {"key1": "value1", "key2": "value2"}
+    mock_yaml = yaml.dump(yaml_content)
+
+    with patch("builtins.open", mock_open(read_data=mock_yaml)):
+        result = _get_config_from_args(mock_args)
+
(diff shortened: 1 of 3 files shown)


GeorgWa (commented):
Did you forget to chek in the CLI testing code?

tests/unit_tests/test_cli.py
```diff
@@ -1,10 +1,119 @@
...
+import yaml
 
-if __name__ == "__main__":
-    unittest.main()
+from alphadia.cli import _get_config_from_args, _get_from_args_or_config, run
+
+# TODO add tests for _get_raw_path_list_from_args_and_config
+
+
+def test_get_config_from_args():
```
GeorgWa, line 12: why don't we get the error, that we can only update existing keys?

tests/unit_tests/test_cli.py
```diff
@@ -1,10 +1,119 @@
...
+
+    # when
+    run()
+
+    mock_search_plan.assert_called_once_with(
+        "/output",
+        {},
+        {
+            # TODO raw_paths missing here
+            "library_path": mock_args.library,
```
GeorgWa, line 84: Maybe I don't understand the mocking yet. why can we check for moch_args.library altough we never initialized it?

tests/unit_tests/test_cli.py
```diff
@@ -1,10 +1,119 @@
...
+@patch("alphadia.cli.SearchPlan")
+def test_cli_minimal_args_all_none(
+    mock_search_plan, mock_init_logging, mock_parse_known_args
+):
+    """Test the run function of the CLI with minimal arguments maps correctly to SearchPlan if nothing given."""
+    mock_args = MagicMock(
+        config=None,
+        version=None,
+        output="/output",
+        fasta=None,
```
GeorgWa, line 104: Same as before. we initialize it but now we don't check for it anymore?

tests/unit_tests/test_cli.py
mschwoer: because this is another dictionary update method 🙈 
I'll add a TODO

tests/unit_tests/test_cli.py
mschwoer: we initialize it as `None` otherwise it will return a mock (cf. above)

GeorgWa (approved):
LGTM

PR #449, opened 2025-01-24
Adjust config and gui for max_iter parameter

anna-charlotte:
Add the parameter `two_step_classifier_max_iterations` to config and GUI. This parameter specifies the number of iterations performed by the two classifiers within the `TwoStepClassifier.fit_predict()` method.

### DIFF
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -38,7 +38,7 @@ class TwoStepClassifier:
         max_iterations : int
             Maximum number of refinement iterations during training.
         train_on_top_n : int
-            Use candidates up to this rank for training.
+            Use candidates up to this rank for training. During inference, all ranks are used.
 
         """
         self.first_classifier = first_classifier
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -17,6 +17,7 @@ class TwoStepClassifier:
         first_fdr_cutoff: float = 0.6,
         second_fdr_cutoff: float = 0.01,
         min_precursors_for_update: int = 5000,
+        max_iterations: int = 5,
         train_on_top_n: int = 1,
     ):
         """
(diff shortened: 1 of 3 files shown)


alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -34,6 +35,10 @@ def __init__(
             The fdr threshold for the second classifier, typically set stricter to ensure high confidence in the final classification results.
         min_precursors_for_update : int, default=5000
             The minimum number of precursors required to update the first classifier.
+        max_iterations : int
+            Maximum number of refinement iterations during training.
+        train_on_top_n : int
+            Use candidates up to this rank for training.
```
GeorgWa, line 41: can you add that all will be used for inference

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -42,6 +47,7 @@ def __init__(
         self.second_fdr_cutoff = second_fdr_cutoff
 
         self.min_precursors_for_update = min_precursors_for_update
+        self._max_iterations = max_iterations
```
GeorgWa, line 50: why is this one private and the others not?

GeorgWa (approved):
LGTM
Just to be a 100% sure. If it says `max_iterations = 1` does this exclude the first iteration we need to pretrain the logistic regression so the two step classifier becomes active at all?

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: The others are private in a newer version, so as soon as you merge the latest changes they should be private

anna-charlotte: see comment here: https://github.com/MannLabs/alphadia/pull/431#discussion_r1923754870

anna-charlotte:
> Just to be a 100% sure. If it says max_iterations = 1 does this exclude the first iteration we need to pretrain the logistic regression so the two step classifier becomes active at all?

In that case the `.first_classifier` will be used only if it has already been fitted, other wise only the `.second_classifier`wille trained and applied, after that the `first_classifier` will be fitted, but only if there are enough precursors found (enough meaning more than `._min_precursors_for_update`

PR #451, opened 2025-01-24
Adjust two-step-classifier files for stricter linting

anna-charlotte:
Adjust the new files that have been added for the two step classifier according to the stricter linter.

### DIFF
--- a/ruff-lint-strict.toml
+++ b/ruff-lint-strict.toml
@@ -6,8 +6,6 @@ select = [
 
 # TODO excluding explicity is not great but it is a workaround for now
 exclude = [
-    "alphadia/fdrx/models/two_step_classifier.py",
-    "alphadia/fdrx/models/logistic_regression.py",
     "alphadia/__*__.py",
     "alphadia/cli.py",
     "alphadia/exceptions.py",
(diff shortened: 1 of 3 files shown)


ruff-lint-strict.toml
```diff
@@ -83,5 +81,7 @@ ignore = [
     "D401", # First line of docstring should be in imperative mood
     "B023", # Function definition does not bind loop variable
     "PD901", # Avoid using the generic variable name `df` for DataFrames"
-    "TCH003" # Move standard library import into a type-checking block
+    "TCH003", # Move standard library import into a type-checking block
+    "PLR0913", # Too many arguments in function definition (only up to 5 are allowed)
```
anna-charlotte, line 85: @mschwoer this error restricts functions to a maximum of five arguments, and for example therefore didn't allow [`TwoStepClassifier._init__()`](https://github.com/MannLabs/alphadia/blob/c2b5c55703802fb722ab3b8f42ea4209245e3c66/alphadia/fdrx/models/two_step_classifier.py#L13). I read that suggested workaround is to introduce a Config class, which would allow us to pass a single config object rather than multiple individual parameters. Do we want to adapt that everywhere where we have more than 5  arguments? Personally I find it more readable with the individual arguments

mschwoer: in this case, you can disable the linting for that specific method by adding `# noqa: PLR0913` .. in this case I would prefer that over a global exclude

mschwoer (approved):
a first step into a new era of code quality :-D

ruff-lint-strict.toml
anna-charlotte: ok! yes, i guess the global exclude kinda goes against the idea of having the stricter linter 😄

PR #452, opened 2025-01-24
restrict num threads

GeorgWa:
Enable restriction of threads on the level of classifier

### DIFF
--- a/alphadia/fdrexperimental.py
+++ b/alphadia/fdrexperimental.py
@@ -1252,8 +1252,7 @@ class BinaryClassifierLegacyNewBatching(Classifier):
 
         x = (x - x.mean(axis=0)) / (x.std(axis=0) + 1e-6)
         self.network.eval()
-        y = np.argmax(self.network(torch.Tensor(x)).detach().numpy(), axis=1)
-        return y
+        return np.argmax(self.network(torch.Tensor(x)).detach().numpy(), axis=1)
 
     @manage_torch_threads(max_threads=2)
     def predict_proba(self, x: np.ndarray):
@@ -1285,8 +1284,7 @@ class BinaryClassifierLegacyNewBatching(Classifier):
 
         x = (x - x.mean(axis=0)) / (x.std(axis=0) + 1e-6)
         self.network.eval()
-        y = self.network(torch.Tensor(x)).detach().numpy()
-        return y
+        return self.network(torch.Tensor(x)).detach().numpy()
 
 
 class FeedForwardNN(nn.Module):
--- a/alphadia/fdrexperimental.py
+++ b/alphadia/fdrexperimental.py
@@ -945,6 +945,8 @@ def get_scaled_training_params(df, base_lr=0.001, max_batch=4096, min_batch=128)
 
 
 class BinaryClassifierLegacyNewBatching(Classifier):
+    max_num_threads = 2
+
     def __init__(
         self,
         input_dim: int = 10,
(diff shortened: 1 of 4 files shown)


alphadia/fdrexperimental.py
```diff
@@ -1041,6 +1045,21 @@ def metrics(self):
...
+            self.is_num_threads_changed = True
+            logger.info(
+                f"Setting torch num_threads to {self.max_num_threads} for FDR classification task"
+            )
+
+    def reset_torch_num_threads(self):
+        if self.is_num_threads_changed:
+            logger.info(f"Resetting torch num_threads to {self.num_threads}")
+            torch.set_num_threads(self.num_threads)
+
```
anna-charlotte, line 1062: Is it intended that `self.if_num_threads_changed` is not set to `False`?

anna-charlotte (approved):

alphadia/fdrexperimental.py
```diff
@@ -1041,6 +1045,21 @@ def metrics(self):
     def metrics(self, metrics):
         self._metrics = metrics
 
+    def restrict_torch_num_threads(self):
```
mschwoer, line 1048: came here to say "this is duplicated logic", but turns out it's duplicated code :-)

Suggestion:
1. Make this a static method (and move it to some `utils` or a `ThreadRestrictor` class..:
```python
def restrict_torch_num_threads(max_num_threads):
    is_num_threads_changed = False
    num_threads = torch.get_num_threads()
    if num_threads > max_num_threads:
        torch.set_num_threads(max_num_threads)
        is_num_threads_changed = True
        logger.info(
            f"Setting torch num_threads to {max_num_threads} for FDR classification task"
        )
    return is_num_threads_changed, num_threads
```
2. similar for `reset_torch_num_threads`
3. use that new method(s) here and in `perform_fdr`

edit: I am beginning to like the decorator idea (cf. below)

alphadia/fdrexperimental.py
```diff
@@ -1246,9 +1269,12 @@ def predict(self, x):
             x.shape[1] == self.input_dim
         ), "Input data must have the same number of features as the fitted classifier."
 
+        self.restrict_torch_num_threads()
```
mschwoer, line 1272: could this be a decorator? then the state (i.e. the number of threads) can be stored transiently in the wrapping method and does not pollute classes..

alphadia/fdrexperimental.py
```diff
@@ -1220,6 +1241,8 @@ def fit(self, x: np.ndarray, y: np.ndarray):
 
         self._fitted = True
 
+        self.reset_torch_num_threads()
```
mschwoer, line 1244: would be more robust if this would be wrapped in `try .. finally` ?

mschwoer (changes requested):
please remove code duplication

alphadia/fdrexperimental.py
GeorgWa: This would be smooth! I agree

alphadia/fdrx/utils.py
```diff
@@ -0,0 +1,51 @@
...
+from collections.abc import Callable
+from typing import Any
+
+import torch
+
+logger = logging.getLogger()
+
+
+def manage_torch_threads(max_threads: int = 2) -> Callable[..., Any]:
+    """Decorator to manage torch thread count during method execution.
```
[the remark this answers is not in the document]
GeorgWa: I made it universal

alphadia/fdrx/utils.py
```diff
@@ -0,0 +1,51 @@
...
+            if original_threads > max_threads:
+                torch.set_num_threads(max_threads)
+                is_threads_changed = True
+                logger.info(
+                    f"Setting torch num_threads to {max_threads} for FDR classification task"
+                )
+
+            try:
+                # Execute the wrapped function
+                return func(*args, **kwargs)
```
mschwoer, line 42: It was never that obvious to me that in case this decorator is used on an instance method, `self` is the first `arg` .. again what learnt ;-)

PR #455, opened 2025-01-28
Fix updating of the `TwoStepClassifier.first_classifier` and add logging

anna-charlotte:
The`TwoStepClassifier.first_classifier` was previously updated based on its performance on a subset following training on that same subset. However, it should be trained on the entire dataset first, and then applied to the subset.

Also added more logging statements to the `TwoStepClassifier`

### DIFF
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -182,26 +182,28 @@ class TwoStepClassifier:
         First extracts the corresponding target/decoy partners from the full dataset
         for each entry in the subset, then uses these pairs to retrain the classifier.
         """
-        df = get_target_decoy_partners(subset_df, full_df)
+        df_train = get_target_decoy_partners(subset_df, full_df)
+        x_train = df_train[x_cols].to_numpy()
+        y_train = df_train[y_col].to_numpy()
 
-        x = df[x_cols].to_numpy()
-        y = df[y_col].to_numpy()
+        df_predict = full_df
+        x_predict = full_df[x_cols].to_numpy()
 
         previous_n_precursors = -1
 
         if self.first_classifier.fitted:
-            df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+            df_predict["proba"] = self.first_classifier.predict_proba(x_predict)[:, 1]
             df_targets = compute_and_filter_q_values(
-                df, self.first_fdr_cutoff, group_columns
+                df_predict, self.first_fdr_cutoff, group_columns
             )
             previous_n_precursors = len(df_targets)
             previous_state_dict = self.first_classifier.to_state_dict()
 
-        self.first_classifier.fit(x, y)
+        self.first_classifier.fit(x_train, y_train)
 
-        df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+        df_predict["proba"] = self.first_classifier.predict_proba(x_predict)[:, 1]
         df_targets = compute_and_filter_q_values(
-            df, self.first_fdr_cutoff, group_columns
+            df_predict, self.first_fdr_cutoff, group_columns
         )
         current_n_precursors = len(df_targets)
 
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -50,11 +50,6 @@ class TwoStepClassifier:
         self.first_fdr_cutoff = first_fdr_cutoff
         self.second_fdr_cutoff = second_fdr_cutoff
 
-        if max_iterations <= 0:
-            raise ValueError(
-                f"`max_iterations` must be greater than 0 (got {max_iterations})"
-            )
-
         self._min_precursors_for_update = min_precursors_for_update
         self._max_iterations = max_iterations
         self._train_on_top_n = train_on_top_n
@@ -111,7 +106,7 @@ class TwoStepClassifier:
                 df_train = self._apply_filtering_with_first_classifier(
                     df, x_cols, group_columns
                 )
-                df_predict = df_train
+                df_predict = df_train  # using the same df for training and predicting, unlike in the following else block.
 
                 previous_target_count_after_first_clf = get_target_count(df_train)
                 self.second_classifier.epochs = 50
@@ -259,16 +254,16 @@ class TwoStepClassifier:
         y_train = df_train[y_col].to_numpy()
 
         x_all = full_df[x_cols].to_numpy()
-        subset_df = full_df[[*group_columns, "decoy"]]
+        reduced_df = full_df[[*group_columns, "decoy"]]
 
         logger.info(f"Fitting first classifier on {len(df_train):,} samples.")
         new_classifier = copy.deepcopy(self.first_classifier)
         new_classifier.fit(x_train, y_train)
 
         logger.info(f"Applying first classifier to {len(x_all):,} samples.")
-        subset_df["proba"] = new_classifier.predict_proba(x_all)[:, 1]
+        reduced_df["proba"] = new_classifier.predict_proba(x_all)[:, 1]
         df_targets = compute_and_filter_q_values(
-            subset_df, self.first_fdr_cutoff, group_columns
+            reduced_df, self.first_fdr_cutoff, group_columns
         )
         n_targets = get_target_count(df_targets)
 
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -148,13 +148,13 @@ class TwoStepClassifier:
                     )
                 )
                 logger.info(
-                    f"{target_count_after_first_clf} precursors found after first classifier, at fdr={self.first_fdr_cutoff}"
+                    f"{target_count_after_first_clf:,} precursors found after first classifier, at fdr={self.first_fdr_cutoff}"
                 )
 
                 if target_count_after_first_clf > previous_target_count_after_first_clf:
                     logger.info(
                         f"Updating the first classifier as new target count increased: "
-                        f"{target_count_after_first_clf} > {previous_target_count_after_first_clf}"
+                        f"{target_count_after_first_clf:,} > {previous_target_count_after_first_clf:,}"
                     )
                     self.first_classifier = new_classifier
                     previous_target_count_after_first_clf = target_count_after_first_clf
@@ -162,7 +162,7 @@ class TwoStepClassifier:
                 else:
                     logger.info(
                         f"Not updating the first classifier as new target count decreased: "
-                        f"{target_count_after_first_clf} < {previous_target_count_after_first_clf}"
+                        f"{target_count_after_first_clf:,} < {previous_target_count_after_first_clf:,}"
                     )
             else:
                 logger.info(
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -186,24 +186,23 @@ class TwoStepClassifier:
         x_train = df_train[x_cols].to_numpy()
         y_train = df_train[y_col].to_numpy()
 
-        df_predict = full_df
-        x_predict = full_df[x_cols].to_numpy()
+        x_all = full_df[x_cols].to_numpy()
 
         previous_n_precursors = -1
 
         if self.first_classifier.fitted:
-            df_predict["proba"] = self.first_classifier.predict_proba(x_predict)[:, 1]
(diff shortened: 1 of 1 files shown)


alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -182,26 +182,27 @@ def _update_first_classifier(
...
 
-        x = df[x_cols].to_numpy()
-        y = df[y_col].to_numpy()
+        x_all = full_df[x_cols].to_numpy()
 
         previous_n_precursors = -1
 
         if self.first_classifier.fitted:
-            df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+            full_df["proba"] = self.first_classifier.predict_proba(x_all)[:, 1]
```
GeorgWa, line 194: Can we reuse the number of precursors at 60% FDR from the first forward pass?
As we have already evaluated the classifier on this subset (full)

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -182,26 +182,27 @@ def _update_first_classifier(
...
+                full_df, self.first_fdr_cutoff, group_columns
             )
             previous_n_precursors = len(df_targets)
             previous_state_dict = self.first_classifier.to_state_dict()
 
-        self.first_classifier.fit(x, y)
+        self.first_classifier.fit(x_train, y_train)
 
-        df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+        full_df["proba"] = self.first_classifier.predict_proba(x_all)[:, 1]
```
GeorgWa, line 203: I think there's still a bug in here 🤔 
Because we are operating on `full_df` in place, `full_df` will have the q-values and proba of the new classifier, independent if the new classifier was better or not. 

I think we should redesign it to make it more clearly separated and less error prone.

GeorgWa: I'm not sure if we actually use these values, but it's a side effect which could lead to issues

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: yes good point, I added a variable `previous_target_count_after_first_clf` to keep track of the count

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: I now changed it as discussed and I extracted only the relevant columns (now `subset_df = full_df[[*group_columns, "decoy"]]`) instead of changing the `full_df` in-place

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -140,15 +182,24 @@ def _preprocess_data(self, df: pd.DataFrame, x_cols: list[str]) -> pd.DataFrame:
```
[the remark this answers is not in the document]
mschwoer: also: I would not bother to answer the bot (sorry, bot)... what I do (just to keep track): I react with 👍 if I changed something, 🙁 if the comment was so-so or if I just decided against it, and 👎 if it's rubbish..

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -49,10 +50,21 @@ def __init__(  # noqa: PLR0913 Too many arguments in function definition (> 5)
         self.first_fdr_cutoff = first_fdr_cutoff
         self.second_fdr_cutoff = second_fdr_cutoff
 
+        if max_iterations <= 0:
```
mschwoer, line 53: could this happen programatically? If not, I am not sure if we need this check

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +95,94 @@ def fit_predict(
...
 
         """
+        logger.info("=== Starting training of TwoStepClassifier ===")
+
         df = self._preprocess_data(df, x_cols)
         best_result = None
-        best_precursor_count = -1
+
+        previous_target_count_after_first_clf = -1
+        previous_target_count_after_second_clf = -1
```
mschwoer, line 104: `clf` = `classifier`?

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +95,94 @@ def fit_predict(
...
         for i in range(self._max_iterations):
+            logger.info(f"Starting iteration {i + 1} / {self._max_iterations}.")
+
+            # extract preselction using first classifier if it is fitted
             if self.first_classifier.fitted and i > 0:
-                df_train, df_predict = self._apply_filtering_with_first_classifier(
+                df_train = self._apply_filtering_with_first_classifier(
                     df, x_cols, group_columns
                 )
+                df_predict = df_train
```
mschwoer, line 114: you just copy the reference, both are pointint to the same `df`, is this on purpose? if so, a comment might help ;-)

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -159,60 +222,57 @@ def _train_and_apply_second_classifier(
...
         """
-        df = get_target_decoy_partners(subset_df, full_df)
+        df_train = get_target_decoy_partners(subset_df, full_df)
+        x_train = df_train[x_cols].to_numpy()
+        y_train = df_train[y_col].to_numpy()
 
-        x = df[x_cols].to_numpy()
-        y = df[y_col].to_numpy()
+        x_all = full_df[x_cols].to_numpy()
+        subset_df = full_df[[*group_columns, "decoy"]]
```
mschwoer, line 262: not sure if "subset" is the correct term here, could be interpreted as "only selected rows" .. maybe "reduced_df`?

Claude's 2 cent:

mschwoer (approved):

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: yessss, tried to shorten those very long variable names...

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: Yes, that's on purpose, we want to use the same df here, but have to specify both `df_train` and `df_predict`, as they are not the same in the following else block. Added a comment!

alphadia/fdrx/models/two_step_classifier.py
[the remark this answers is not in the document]
anna-charlotte: haha ok :D

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: oh and also just saw that I was overwriting the given `subset_df`.. Renamed it to `reduced_df`

alphadia/fdrx/models/two_step_classifier.py
mschwoer: with "programatically" I meant indeed the concept of "user exposure" ..

even if it was exposed, I would do such checks upfront in a centralized "config check" component that we are _planning to implement soon (TM)_ :-) @GeorgWa FYI

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
...
-                    f"due to decreased target count ({current_target_count} < {best_precursor_count})"
+                    f"Training stopped on iteration {i + 1}. Decrease in target count from "
+                    f"{previous_target_count_after_second_clf:,} to {current_target_count:,}."
                 )
                 return best_result
 
-            best_precursor_count = current_target_count
-            best_result = predictions
+            previous_target_count_after_second_clf = current_target_count
+            best_result = df_after_second_clf
```
GeorgWa, line 136: We are keeping this to always return the best results?
I think if we decide to go without multiple iterations we should think about removing this for memory reasons :)

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
...
+
+            # extract preselction using first classifier if it is fitted
             if self.first_classifier.fitted and i > 0:
-                df_train, df_predict = self._apply_filtering_with_first_classifier(
+                df_train = self._apply_filtering_with_first_classifier(
                     df, x_cols, group_columns
                 )
+                df_predict = df_train  # using the same df for training and predicting, unlike in the following else block.
+
+                previous_target_count_after_first_clf = get_target_count(df_train)
```
GeorgWa, line 111: Should we add a log statement like "applying linear classifier at 60% FDR - xxx targets and xxx decoys"?

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
...
+                f"{current_target_count:,} precursors found after second classifier, "
+                f"at fdr={self.second_fdr_cutoff}"
+            )
 
-            # Update first classifier if enough confident predictions
+            # update first classifier if enough confident predictions
             if current_target_count > self._min_precursors_for_update:
-                self._update_first_classifier(
-                    df_filtered, df, x_cols, y_col, group_columns
+                logger.info(
```
GeorgWa, line 145: I think we can skipp this statement and only log if it wasnt the case. This will usually be true.

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
...
+                    )
                 )
+                logger.info(
+                    f"{target_count_after_first_clf:,} precursors found after first classifier, "
+                    f"at fdr={self.first_fdr_cutoff}"
+                )
+
+                if target_count_after_first_clf > previous_target_count_after_first_clf:
+                    logger.info(
+                        f"Updating the first classifier as new target count increased: "
```
GeorgWa, line 161: Maybe "target count" is not very informative to the user who doesn't know about the implementation.
Maybe something like "previous version xxx precursors, current version xxx precursors - updating classifier"

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
...
+                        f"{target_count_after_first_clf:,} > {previous_target_count_after_first_clf:,}"
+                    )
+                    self.first_classifier = new_classifier
+                    previous_target_count_after_first_clf = target_count_after_first_clf
+
+                else:
+                    logger.info(
+                        f"Not updating the first classifier as new target count decreased: "
```
GeorgWa, line 169: same log statement as before but without update

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -140,15 +189,24 @@ def _preprocess_data(self, df: pd.DataFrame, x_cols: list[str]) -> pd.DataFrame:
 
     def _apply_filtering_with_first_classifier(
         self, df: pd.DataFrame, x_cols: list[str], group_columns: list[str]
-    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+    ) -> pd.DataFrame:
         """Apply first classifier to filter data for the training of the second classifier."""
+        n_precursors = get_target_count(df)
```
GeorgWa, line 194: I feel like we could skip these logging statements

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -159,60 +217,57 @@ def _train_and_apply_second_classifier(
         group_columns: list[str],
     ) -> pd.DataFrame:
         """Train second_classifier and apply it to get predictions."""
+        logger.info(
```
GeorgWa, line 220: same about these logging statements

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -83,52 +90,94 @@ def fit_predict(
             DataFrame containing predictions and q-values
 
         """
+        logger.info("=== Starting training of TwoStepClassifier ===")
+
         df = self._preprocess_data(df, x_cols)
         best_result = None
-        best_precursor_count = -1
+
+        previous_target_count_after_first_clf = -1
```
GeorgWa, line 98: can you add a comment that they will track precursor and at which FDR (variable name)

GeorgWa (approved):
Much cleaner implementation and naming! I like it 😄 

Sorry that I'm a bit annoying about the logging :D I'm thinking about the minimal number of lines we need to characterize what's happening without overloading the user.

1. Target's and decoys after first classifier 
2. Targets after second classifier @1% FDR
3. Comparison old and new classifier and update decision

If you don't want to kick out the other log statements we could also add them as `logger.debug` statements with higher verbosity.

anna-charlotte (commented):
I removed some of the logging statements, and moved others to `logging.debug` as suggested

PR #457, opened 2025-01-28
UI fixes

mschwoer:
Fix blocking bug that UI does not work with new config implementation.

Plus some smaller improvements on they way.

(see individual commits)

### DIFF
--- a/alphadia/exceptions.py
+++ b/alphadia/exceptions.py
@@ -73,22 +73,29 @@ class ConfigError(BusinessError):
     _key = ""
     _config_name = ""
 
-    def __init__(self, key: str, config_name: str):
+    def __init__(self, key: str, value: str, config_name: str):
         self._key = key
+        self._value = value
         self._config_name = config_name
 
 
 class KeyAddedConfigError(ConfigError):
     """Raise when a key should be added to a config."""
 
-    def __init__(self, key: str, config_name: str):
-        super().__init__(key, config_name)
-        self._detail_msg = f"Defining new keys is not allowed when updating a config: key='{self._key}', config_name='{self._config_name}'"
+    def __init__(self, key: str, value: str, config_name: str):
+        super().__init__(key, value, config_name)
+        self._detail_msg = (
+            f"Defining new keys is not allowed when updating a config: "
+            f"key='{self._key}', value='{self._value}', config_name='{self._config_name}'"
+        )
 
 
 class TypeMismatchConfigError(ConfigError):
     """Raise when the type of a value does not match the default type."""
 
-    def __init__(self, key: str, config_name: str, extra_msg: str):
-        super().__init__(key, config_name)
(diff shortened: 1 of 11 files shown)


alphadia/exceptions.py
```diff
@@ -64,6 +72,19 @@ class NotDiaDataError(BusinessError):
     _msg = "Could not find cycle shape. Please check if this is a valid DIA data set."
 
 
+class NoLibraryAvailableError(UserError):
+    """Raise when no ;ibrary is available."""
```
GeorgWa, line 76: Nit ;library

alphadia/exceptions.py
```diff
@@ -64,6 +72,19 @@ class NotDiaDataError(BusinessError):
...
+
+    _error_code = "NO_LIBRARY_AVAILABLE"
```
GeorgWa, line 78: I think we have to rephrase the message a bit to make it appeal the mind of proteomics users. Fully predicted library library searches are called “library free” search by many. 

If a user wants to predict the library from fasts and gets the message “no library loaded” he might be confused because he never wanted to use a library in the first place :D

alphadia/search_step.py
```diff
@@ -271,7 +271,7 @@ def get_run_data(self) -> Generator[tuple[str, str, SpecLibFlat]]:
         """Generator for raw data and spectral library."""
 
         if self.spectral_library is None:
-            raise ValueError("no spectral library loaded")
+            raise NoLibraryAvailableError()
```
GeorgWa, line 274: If we raise here, the error message should say that there needs to be a library or library prediction as mentioned. 

Otherwise, we could also raise on the level of the search plan with individual errors.

docs/installation.md
```diff
@@ -108,6 +108,14 @@ The GUI can then be started by typing
 npm run dev
 ```
 
+Ignore the error message telling you that the `BundledExecutionEngine` is not available
```
GeorgWa, line 111: Should I remove this error message?

gui/src/renderer/pages/Home.js
```diff
@@ -55,7 +55,7 @@ const VersionCard = () => {
         let currentActiveVersion = profile.executionEngines[profile.activeIdx].versions.filter((versionEntry) => versionEntry.name === "alphadia")[0].version
         versionOutput = currentActiveVersion
 
-        if (currentActiveVersion != "1.2.0") {
+        if (currentActiveVersion != "1.2.0") { // TODO fix
```
GeorgWa, line 58: I wouldn’t be surprised if Claude could figure this out in a single prompt 😆

GeorgWa (approved):
LGTM

alphadia/exceptions.py
mschwoer: changed it to "available"

PR #460, opened 2025-01-29
Add parameter to config and gui to set the min precursors required for update

anna-charlotte:
Adding the `two_step_classifier_min_precursors_for_update` parameter to config and gui (experimental)

### DIFF
--- a/alphadia/fdrx/models/two_step_classifier.py
+++ b/alphadia/fdrx/models/two_step_classifier.py
@@ -182,26 +182,28 @@ class TwoStepClassifier:
         First extracts the corresponding target/decoy partners from the full dataset
         for each entry in the subset, then uses these pairs to retrain the classifier.
         """
-        df = get_target_decoy_partners(subset_df, full_df)
+        df_train = get_target_decoy_partners(subset_df, full_df)
+        x_train = df_train[x_cols].to_numpy()
+        y_train = df_train[y_col].to_numpy()
 
-        x = df[x_cols].to_numpy()
-        y = df[y_col].to_numpy()
+        df_predict = full_df
+        x_predict = full_df[x_cols].to_numpy()
 
         previous_n_precursors = -1
 
         if self.first_classifier.fitted:
-            df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+            df_predict["proba"] = self.first_classifier.predict_proba(x_predict)[:, 1]
             df_targets = compute_and_filter_q_values(
-                df, self.first_fdr_cutoff, group_columns
+                df_predict, self.first_fdr_cutoff, group_columns
             )
             previous_n_precursors = len(df_targets)
             previous_state_dict = self.first_classifier.to_state_dict()
 
-        self.first_classifier.fit(x, y)
+        self.first_classifier.fit(x_train, y_train)
 
-        df["proba"] = self.first_classifier.predict_proba(x)[:, 1]
+        df_predict["proba"] = self.first_classifier.predict_proba(x_predict)[:, 1]
         df_targets = compute_and_filter_q_values(
-            df, self.first_fdr_cutoff, group_columns
+            df_predict, self.first_fdr_cutoff, group_columns
         )
         current_n_precursors = len(df_targets)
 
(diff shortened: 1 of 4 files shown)


alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -90,91 +86,63 @@ def fit_predict(
             DataFrame containing predictions and q-values
 
         """
+        min_train_size = 1
         logger.info("=== Starting training of TwoStepClassifier ===")
 
         df = self._preprocess_data(df, x_cols)
-        best_result = None
+        df_train = df[df["rank"] < self._train_on_top_n]
```
mschwoer, line 93: just from the name, I would assume `_train_on_top_n` is a boolean, but then this line would make no sense .. maybe find a better name?

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -90,91 +86,63 @@ def fit_predict(
...
-                )
-
-                previous_target_count_after_first_clf = get_target_count(df_train)
-                self.second_classifier.epochs = 50
-            else:
-                logger.debug("First classifier not fitted yet. Proceeding without it.")
-                df_train = df[df["rank"] < self._train_on_top_n]
-                df_predict = df
+        # train and apply NN classifier
+        self.second_classifier.epochs = 10
```
mschwoer, line 97: deliberately hardcoded?

alphadia/fdrx/models/two_step_classifier.py
```diff
@@ -288,12 +287,30 @@ def get_target_count(df: pd.DataFrame) -> int:
...
+    group_columns: list[str] | None = None,
+    qval_col: str = "qval",
+    scale_by_target_decoy_ratio: bool = True,  # noqa: FBT001, FBT002
 ) -> pd.DataFrame:
     """Compute q-values for each entry after keeping only best entries per group."""
+    scaling_factor = 1.0
+    if scale_by_target_decoy_ratio:
+        n_targets = (df["decoy"] == 0).sum()
+        n_decoys = (df["decoy"] == 1).sum()
+        scaling_factor = round(n_targets / n_decoys, 3)
```
mschwoer, line 300: please avoid raising `ZeroDivisionError` here (either by catching it or by adding a small epsilon to the denominator (not sure how the latter will affect the `isfinite` check though)

alphadia/workflow/manager.py
```diff
@@ -808,22 +810,21 @@ def load_classifier_store(self, path: None | str = None):
...
-                if file.endswith(".pth"):
-                    classifier_hash = file.split(".")[0]
-
-                    if classifier_hash not in self.classifier_store:
-                        classifier = deepcopy(self.classifier_base)
-                        classifier.from_state_dict(
-                            torch.load(os.path.join(path, file), weights_only=False)
-                        )
-                        self.classifier_store[classifier_hash].append(classifier)
+        for file in os.listdir(path):
```
mschwoer, line 818: https://www.stuartellis.name/articles/python-modern-practices/#use-osscandir-instead-of-oslistdir 
Would not have expected reading that article yesterday would come in handy so quickly ;-)

alphadia/fdrx/models/two_step_classifier.py
anna-charlotte: Yes, but the value was chosen somewhat arbitrarily.  The reason I set it to 10 was to avoid the error where `BinaryClassifierLegacyNewBatching.fit()` crashes in [`model_selection.train_test_split(x, y, test_size=self.test_size)`](https://github.com/MannLabs/alphadia/blob/ab68abb4ca429c6f51be993018dfdba9f39fa57a/alphadia/fdrexperimental.py#L360) when there aren’t enough samples in x for splitting. 10 worked for me, but I agree it’s not ideal and as I said, chosen a bot by random. Do you have a suggestion on what do do instead maybe?

mschwoer: at least, make it either a module-wide constant or create a new method parameter and set it as default (then is is more obvious that there is a knob to tune)..
if it makes sense to have the user tune it -> config

anna-charlotte: oops sorry, I mixed something up here, and was talking about something else. Yes, this one is deliberately set to 10, but I will add a `#TODO` to line 126 where we set it `= 50`

PR #463, opened 2025-01-30
fix tests by adapting to changes in pytorch 2.6

anna-charlotte:
In [PyTorch 2.6](https://pytorch.org/blog/pytorch2-6/), the default value of `weights_only` in `torch.load(path, weights_only=True)` was changed to `True`, which caused our tests to fail. Their intention is that this is more secure, because not an entire model object is loaded by default, but instead only the weights. 
As a quick fix, I changed our use-cases to `weights_only=False`.

### DIFF
--- a/alphadia/workflow/manager.py
+++ b/alphadia/workflow/manager.py
@@ -820,7 +820,9 @@ class FDRManager(BaseManager):
 
                     if classifier_hash not in self.classifier_store:
                         classifier = deepcopy(self.classifier_base)
-                        classifier.from_state_dict(torch.load(os.path.join(path, file)))
+                        classifier.from_state_dict(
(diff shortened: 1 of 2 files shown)


alphadia/workflow/manager.py
```diff
@@ -820,7 +820,9 @@ def load_classifier_store(self, path: None | str = None):
 
                     if classifier_hash not in self.classifier_store:
                         classifier = deepcopy(self.classifier_base)
-                        classifier.from_state_dict(torch.load(os.path.join(path, file)))
+                        classifier.from_state_dict(
```
mschwoer, line 823: thank you for fixing this ..

@GeorgWa this brings me back to our discussion to pin also the transient variables ..

mschwoer (approved):

PR #466, opened 2025-02-10
Config simplifications

mschwoer:
Small tweaks to align UI and config (i.e. avoid that config will show a diff even though no UI parameter has been changed)

For both keys, I have questions @GeorgWa:
- `peptdeep_model_type`: why is `None` and `generic` treated differently in code although docs say they mean the same
- `channel_filter`: UI docs say "Leave empty to load all available channels.". Is this the behaviour of `0`?

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -96,7 +96,7 @@ custom_modifications:
     composition: C(12)
 
 search:
-  channel_filter: '0'
+  channel_filter: ""
   exclude_shared_ions: True
   compete_for_fragments: True
 
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -250,6 +250,7 @@ fdr:
 search_output:
   peptide_level_lfq: false
   precursor_level_lfq: false
+  save_fragment_quant_matrix: false #advanced feature to write out quantitative matrix with fragment ion intensities e.g. for AlphaQuant
   min_k_fragments: 12
   min_correlation: 0.9
   num_samples_quadratic: 50
@@ -257,6 +258,7 @@ search_output:
   normalize_lfq: True
   # can be either "parquet" or "tsv"
   file_format: "tsv"
+  file_format_advanced: "parquet"
 
 # Configuration for the optimization of search parameters. These parameters should not normally be adjusted and are for the use of experienced users only.
 optimization:
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -96,7 +96,7 @@ custom_modifications:
     composition: C(12)
 
 search:
-  channel_filter: ""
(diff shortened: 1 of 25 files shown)


alphadia/constants/default.yaml
```diff
@@ -96,7 +96,7 @@ custom_modifications:
     composition: C(12)
 
 search:
-  channel_filter: '0'
+  channel_filter: "" # will be mapped to "0" -> load all channels
```
GeorgWa, line 99: That's not 100% right. `channel_filter: ""` will only load the default unmultiplexed channel which is the intended behaviour. So we just need to change the comment.

alphadia/constants/default.yaml
```diff
@@ -61,8 +61,8 @@ library_prediction:
   # set path for custom peptdeep model. If set to null, the default model will be used
   peptdeep_model_path: null
 
-  # set peptdeep model type. Possible values are 'generic', 'phospho', 'digly'. If set to null, the generic model will be used
-  peptdeep_model_type: null
+  # set peptdeep model type. Possible values are 'generic', 'phospho', 'digly'.
+  peptdeep_model_type: 'generic'
```
GeorgWa, line 65: 👍🏻

GeorgWa (approved):
LGTM

alphadia/libtransform.py
```diff
@@ -327,11 +327,13 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
         model_mgr = ModelManager(device=device)
 
         # will load other model than default generic
-        if self.peptdeep_model_type:
+        if (
+            self.peptdeep_model_type
+        ):  # TODO: why is this not required if peptdeep_model_type == None?
```
mschwoer, line 332: @GeorgWa ? `generic` and `None` are treated differently although docstring days its equvalent?

GeorgWa: It's a bit ugly. If `peptdeep_model_type = None` it's left for peptdeep to decide what model to load which is generic. The idea is that we can upgrade the peptdeep model default to `generic-v2` and alphadia doens't need to know. Let me know what you prefer.

GeorgWa (approved):

alphadia/libtransform.py
mschwoer: okay, changed the docstring..
still puzzling me: in case `peptdeep_model_type=None`, `model_mgr.load_installed_models(self.peptdeep_model_type)` is not called at all? is this desired?

GeorgWa: It's called by default in the ModelManager on init:
https://github.com/MannLabs/alphapeptdeep/blob/c5a05e17e6449504c459fcc53f74e25ed5545068/peptdeep/pretrained_models.py#L294C15-L294C37

In case we want something else than the default we are calling it again with `"peptdeep_model_type"`. This is why `"generic"` and `""` behave the same.

PR #469, opened 2025-02-17
Ci improvements

mschwoer:
some minor improvements on the CI scripts

### DIFF
--- a/tests/run_unit_tests.sh
+++ b/tests/run_unit_tests.sh
@@ -4,7 +4,7 @@ ENV_NAME=alphadia
 
 INCLUDE_SLOW_TESTS=${1:-false}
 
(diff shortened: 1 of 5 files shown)


tests/run_unit_tests.sh
```diff
@@ -1,10 +1,9 @@
 #!/bin/bash
 
-ENV_NAME=alphadia
+ENV_NAME=${1:-alphadia}
```
GeorgWa, line 3: should we set the default to `alphadia-env`?

GeorgWa (approved):
LGTM. A small summary of the goals of the improvment would've helped with review :)

tests/run_unit_tests.sh
mschwoer: at first I thought so too, but probably many people call their local env `alphadia` ;-)

PR #470, opened 2025-02-17
Improve multistep docs

mschwoer:
Improving documentation of multistep workflow.
Note: did deliberately not turn around the guide, as this would mean a major rework and a lot of duplication.


GeorgWa (approved):
How should we handle this guide if we have multistep and the GUI. I feel like we should maybe separate the multistep search into a separate guide and leave this (with the old version)?

mschwoer:
Isn't the "we have multistep and the GUI" part exactly what is covered here? https://github.com/MannLabs/alphadia/blob/499701f3c0f677a362ff69ef2280ed78c8b7c7bb/docs/guides/transfer-dimethyl.md#integrated-multistep-workflow

I agree it's not 100% streamlined (as it is pointing to the first part of the docs some times), but I feel these two should be quite closely togehter (as the 'manual' part gives a lot of intuition about the parameters, which on the other hand I don't want to copy)

PR #471, opened 2025-02-18
Highlight non default

mschwoer:
Highlight non-default paramters in UI.

Unrelated: some cosmetics around the workflow selection

image

Idea was conceived after drawing red boxes for 15 minutes yesterday :-D


GeorgWa (approved):
Frontend Magnus is at it :D

PR #475, opened 2025-02-18
Change requirements handling

mschwoer:
Changing handling of requirements:
- see the Readme here: https://github.com/MannLabs/alphashared/pull/22
- cf. also this idea https://github.com/MannLabs/alphadia/pull/468
- move neptune to test dependenceis -> saves quite some packages: https://github.com/MannLabs/alphadia/pull/480

@GeorgWa please check the changes to `requirements.txt`, maybe some constraints need to be re-added?


requirements/requirements.txt
```diff
@@ -1,23 +1,20 @@
...
-alphabase==1.5.0 # test: tolerate_version
-peptdeep==1.3.0 # test: tolerate_version
-progressbar==2.5
-neptune==1.10.4
-seaborn==0.13.2
-rocket_fft==0.2.5
-xxhash==3.4.1
-torchmetrics==1.4.0.post0
-transformers==4.40.2
+alphabase==1.5.0
```
GeorgWa, line 6: Shouldn't this be the requirements from the old `requirements_loose.txt`?
E.g. the minimal known constraints for feature compatability.
```
alpharaw>=0.3.1
alphatims>=1.0.8
alphabase>=1.5.0
peptdeep>=1.3.0
directlfq>=0.2.19
transformers<=4.40.2
numpy<2
scipy>=1.12.0
```

mschwoer: agree .. note that I dropped the constraint on `transformers` as we do support the latest version now

requirements/_requirements.freeze.txt
```diff
@@ -0,0 +1,122 @@
...
+nvidia-cuda-runtime-cu12==12.4.127
+nvidia-cudnn-cu12==9.1.0.70
+nvidia-cufft-cu12==11.2.1.3
+nvidia-curand-cu12==10.3.5.147
+nvidia-cusolver-cu12==11.6.1.9
+nvidia-cusparse-cu12==12.3.1.170
+nvidia-cusparselt-cu12==0.6.2
+nvidia-nccl-cu12==2.21.5
+nvidia-nvjitlink-cu12==12.4.127
+nvidia-nvtx-cu12==12.4.127
```
mschwoer, line 65: Note that these seem to be linux-specific, cf. https://github.com/MannLabs/alphadia/actions/runs/13411418109/job/37462368960#step:10:141
(same with triton)

We could add such dependencies to some "blacklist"?
Then the process would be:
1. upgrade a dependency
2. run freeze workflow
3. run full test matrix
4. in case of error: update blacklist and repeat from 2.?

@GeorgWa ?

mschwoer: would look like thid https://github.com/MannLabs/alphadia/pull/484/files

mschwoer: the issues here https://github.com/MannLabs/alphapeptdeep/pull/227#issuecomment-2658034516 requires another mechanism: being able to add dependencies explicitly as constraints

https://github.com/MannLabs/alphadia/pull/475/commits/01acfed66556edd498831a05dfefd06ddc29207f

requirements/requirements_development.txt
```diff
@@ -2,13 +2,12 @@
...
 bump-my-version
 pipdeptree
 ipykernel
-psutil
 
 pre-commit==3.7.0
-pywin32; sys_platform=='win32'
+#pywin32; sys_platform=='win32'
+
+neptune
```
GeorgWa, line 13: I think we could drop neptune from dev and only install it in the GH action where we actually use it. Might be an antipattern though.

GeorgWa (approved):
LGTM

requirements/requirements_development.txt
mschwoer: *cough* https://github.com/MannLabs/alphadia/pull/475/commits/af244ffd87b820a30fd014deb4ffe27d09778c12

PR #479, opened 2025-02-18
Simplify config file and GUI

GeorgWa:
Review only after #466 is merged :D 
- Some simplifications of the config yaml. Small config sections where merged with larger ones.
- Improved GUI by showing all config sections but hiding the parameters in an advanced section.
Screenshot 2025-02-18 at 21 32 08

edit:
breaking: config "search_advanced":"top_k_fragments" -> "search":"top_k_fragments"


alphadia/constants/default.yaml
```diff
@@ -10,13 +10,27 @@ fasta_paths: []
...
 
 general:
+
   thread_count: 10
-  # maximum number of threads or processes to use per raw file
+  # Enables separate transfer learning step
+  transfer_step_enabled: False
+
+  # Enables separate MBR step
+  mbr_step_enabled: False
```
mschwoer, line 19: this is the only breaking change, right? (which is not really breaking as these two are not part of an official release yet)

mschwoer: update: there's "search_advanced":"top_k_fragments" -> "search":"top_k_fragments"

mschwoer (approved):
LGTM! new UI looks quite clean

gui/src/renderer/components/ParameterGroup.js
```diff
@@ -26,43 +22,62 @@ const ParameterGroup = ({parameterGroup, sx}) => {
...
-                            <Grid item xs={12}>
-                                <ParameterInput
-                                    parameter = {parameter}
-                                    parameter_group_id = {parameterGroup.id}
-                                    onChange = {(value) => {dispatch(
-                                        {type: 'updateParameter',
-                                        parameterGroupId: parameterGroup.id,
-                                        parameterId: parameter.id,
-                                        value: value
+                        <Collapse in={advancedOpen} sx={{width: "100%"}}>
```
mschwoer, line 62: maybe we could add an "uncollapse all" for the "advanced users"?

GeorgWa: It's a good suggestion but I feel it's not super important.

PR #492, opened 2025-02-20
Save advanced parameters from GUI

GeorgWa:
Solves a bug where advanced parameters were not saved to config.yaml

### DIFF
--- a/gui/src/main/modules/workflows.js
+++ b/gui/src/main/modules/workflows.js
@@ -99,6 +99,9 @@ function workflowToConfig(workflow) {
         config.parameters.forEach((parameter) => {
             output[config.id][parameter.id] = parameter.value
         })
+        config.parameters_advanced.forEach((parameter) => {
+            output[config.id][parameter.id] = parameter.value
+        })
     })
 
     return output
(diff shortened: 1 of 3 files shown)


gui/src/main/modules/workflows.js
```diff
@@ -99,6 +99,9 @@ function workflowToConfig(workflow) {
         config.parameters.forEach((parameter) => {
             output[config.id][parameter.id] = parameter.value
         })
+        config.parameters_advanced.forEach((parameter) => {
+            output[config.id][parameter.id] = parameter.value
+        })
```
mschwoer, line 104: why do we need this distinction anyway? could we not just treat all parameters on the same level?
("advanced" is just display logic, no?)

mschwoer (approved):

gui/src/renderer/logic/context.js
```diff
@@ -63,12 +60,27 @@ export function methodReducer(method, action) {
 
             return {...method, config: new_config}
 
+        case 'updateParameterAdvanced':
```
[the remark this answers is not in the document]
GeorgWa: Good point!

gui/src/main/modules/workflows.js
GeorgWa: Yes, this might actually be more elegant. I choose this route as it was simpler to implement with fewer code changes.

PR #495, opened 2025-02-27
Update prediction config to use enabled pattern

GeorgWa:
So far, all but one config groups have used the following pattern for configuration:
```yaml
config_group:
   enabled: true
```
This PR introduces a breaking change and updates the config group library prediction to use the same pattern.


mschwoer (approved):
LGTM, tests need to be fixed though

PR #499, opened 2025-03-11
Small tweaks

mschwoer:
Some small improvements that makes life easier.

### DIFF
--- a/alphadia/exceptions.py
+++ b/alphadia/exceptions.py
@@ -40,6 +40,16 @@ class UserError(CustomError):
     """
 
 
+class GenericUserError(UserError):
+    """Raise when something is wrong with the user."""
+
+    _error_code = "USER_ERROR"
+
+    def __init__(self, msg: str, detail_msg: str = ""):
+        self._msg = msg
+        self._detail_msg = detail_msg
+
+
 class NoPsmFoundError(BusinessError):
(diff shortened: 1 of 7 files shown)


alphadia/exceptions.py
```diff
@@ -40,6 +40,16 @@ class UserError(CustomError):
     """
 
 
+class GenericUserError(UserError):
+    """Raise when something is wrong with the user."""
```
GeorgWa, line 44: When something is wrong with the user or the user input/configuration? 😆

alphadia/workflow/config.py
```diff
@@ -135,6 +135,7 @@ def _update(
     update_config: dict,
     tracking_dict: dict,
     config_name: str,
+    parent_keys: str = "",
```
GeorgWa, line 138: does this class need to know that the parent keys are being used?
Maybe somehting like "expection_location_str" or such?

GeorgWa (approved):
LGTM

alphadia/exceptions.py
mschwoer: I think the latter is a subset of the former.

alphadia/workflow/config.py
mschwoer: well, this class knows it's recursive .. so the urge to know its parents is understandable I guess

PR #500, opened 2025-03-11
Frag feature extraction

vuductung:
This pull request includes changes to the `alphadia/outputtransform.py` file to enhance the flexibility and efficiency of data processing, as well as the addition of a new Jupyter notebook for feature extraction.

Improvements to data processing:

* [`alphadia/outputtransform.py`](diffhunk://#diff-f0e8190c13abf4186b0a22f91535423e70c59b351d545f4bc3a0346fcb2bb14bL72-R79): Modified the `QuantBuilder` class to accept a list of columns instead of a single column, allowing for more flexible data processing.
* [`alphadia/outputtransform.py`](diffhunk://#diff-f0e8190c13abf4186b0a22f91535423e70c59b351d545f4bc3a0346fcb2bb14bL128-R170): Updated the `accumulate_frag_df` method to handle multiple columns dynamically and return a list of dataframes, improving the method's scalability.
* [`alphadia/outputtransform.py`](diffhunk://#diff-f0e8190c13abf4186b0a22f91535423e70c59b351d545f4bc3a0346fcb2bb14bL299-R303): Adjusted the `prepare_df` function to return all specified columns, enhancing the data preparation step.
* [`alphadia/outputtransform.py`](diffhunk://#diff-f0e8190c13abf4186b0a22f91535423e70c59b351d545f4bc3a0346fcb2bb14bL797-R803): Modified the `build_lfq_tables` function to work with the new list of dataframes returned by `accumulate_frag_df`, ensuring compatibility with the updated data processing logic.

New feature extraction notebook:

* [`nbs/feature_extraction/feature_extraction.ipynb`](diffhunk://#diff-e763816b35d76062e2886dda3f1793a289bf8d80b05b88289b67d191bd32f655R1-R165): Added a new Jupyter notebook for feature extraction, demonstrating how to use the updated `QuantBuilder` class and save the extracted features to CSV files.

### DIFF
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -70,13 +70,7 @@ def get_frag_df_generator(folder_list: list[str]):
 
 class QuantBuilder:
     def __init__(self, psm_df, columns=['intensity',
-                                        'mass_error',
-                                        'correlation',
-                                        'height',
-                                        'charge',
-                                        'mz_observed',
-                                        'type',
-                                        'number']):
+                                        'correlation']):
         self.psm_df = psm_df
         self.columns = columns
 
@@ -88,7 +82,7 @@ class QuantBuilder:
         Parameters
         ----------
 
-        folder_list: List[str]
+        folder_list: List[str],
             List of folders containing the frag.tsv file
 
         Returns
@@ -297,7 +291,7 @@ class QuantBuilder:
         return protein_df
 
 
-def prepare_df(df, psm_df, columns=["intensity"]):
+def prepare_df(df, psm_df, columns=["intensity", "correlation"]):
     df = df[df["precursor_idx"].isin(psm_df["precursor_idx"])].copy()
     df["ion"] = utils.ion_hash(
         df["precursor_idx"].values,
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -150,12 +150,25 @@ class QuantBuilder:
                 df_list[idx].rename(columns={col: raw_name}, inplace=True)
 
         for idx in range(len(df_list)):
-            df_list[idx] = self._data_prep(df_list[idx])
+            df_list[idx] = self._add_annotation(df_list[idx])
             
         return df_list
     
-    def _data_prep(self, df):
+    def _add_annotation(self, df: pd.DataFrame) -> pd.DataFrame:
 
+        """Add annotation to the fragment data, including protein group, mod_seq_hash, mod_seq_charge_hash
+
+        Parameters
+        ----------
+        df: pd.DataFrame
+            Fragment data
+
+        Returns
+        -------
+        pd.DataFrame
+            Fragment data with annotation
+        """
+        
         df.fillna(0, inplace=True)
 
         df["precursor_idx"] = df["precursor_idx"].astype(np.uint32)
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -69,7 +69,7 @@ def get_frag_df_generator(folder_list: list[str]):
 
 
 class QuantBuilder:
-    def __init__(self, psm_df, column=['intensity',
+    def __init__(self, psm_df, columns=['intensity',
                                         'mass_error',
                                         'correlation',
                                         'height',
@@ -78,7 +78,7 @@ class QuantBuilder:
                                         'type',
                                         'number']):
         self.psm_df = psm_df
-        self.column = column
+        self.columns = columns
 
     def accumulate_frag_df_from_folders(
         self, folder_list: list[str]
@@ -130,18 +130,18 @@ class QuantBuilder:
             logger.warning(f"no frag file found for {raw_name}")
             return None
 
-        df = prepare_df(df, self.psm_df, column=self.column)
+        df = prepare_df(df, self.psm_df, column=self.columns)
 
         df_list = []
-        for col in self.column:
+        for col in self.columns:
             feat_df = df[["precursor_idx", "ion", col]].copy()
             feat_df.rename(columns={col: raw_name}, inplace=True)
             df_list.append(feat_df)
         
         for raw_name, df in df_iterable:
-            df = prepare_df(df, self.psm_df, column=self.column)
+            df = prepare_df(df, self.psm_df, column=self.columns)
             
-            for idx, col in enumerate(self.column):
+            for idx, col in enumerate(self.columns):
                 df_list[idx] = df_list[idx].merge(
                     df[["ion", col, "precursor_idx"]],
                     on=["ion", "precursor_idx"],
--- a/alphadia/outputtransform.py
+++ b/alphadia/outputtransform.py
@@ -149,11 +149,8 @@ class QuantBuilder:
                 )
                 df_list[idx].rename(columns={col: raw_name}, inplace=True)
 
-        for idx in range(len(df_list)):
-            df_list[idx] = self._add_annotation(df_list[idx])
-            
-        return df_list
-    
+        return [ self._add_annotation(df) for df in df_list]
+                
     def _add_annotation(self, df: pd.DataFrame) -> pd.DataFrame:
 
         """Add annotation to the fragment data, including protein group, mod_seq_hash, mod_seq_charge_hash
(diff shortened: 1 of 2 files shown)


alphadia/outputtransform.py
```diff
@@ -69,7 +69,14 @@ def get_frag_df_generator(folder_list: list[str]):
 
 
 class QuantBuilder:
-    def __init__(self, psm_df, column="intensity"):
+    def __init__(self, psm_df, column=['intensity',
```
mschwoer, line 72: please rename `column` => `columns`

alphadia/outputtransform.py
```diff
@@ -125,45 +132,42 @@ def accumulate_frag_df(
...
-                on=["ion", "precursor_idx"],
-                how="outer",
-            )
-            quality_df.rename(columns={"correlation": raw_name}, inplace=True)
+        for idx in range(len(df_list)):
+            df_list[idx] = self._data_prep(df_list[idx])
+            
+        return df_list
+    
+    def _data_prep(self, df):
```
mschwoer, line 157: this method could deserve a little more love (=name, type hints, docstring, ..)

alphadia/outputtransform.py
```diff
@@ -125,45 +132,42 @@ def accumulate_frag_df(
...
-        quality_df.rename(columns={"correlation": raw_name}, inplace=True)
-
+        df_list = []
+        for col in self.column:
+            feat_df = df[["precursor_idx", "ion", col]].copy()
+            feat_df.rename(columns={col: raw_name}, inplace=True)
+            df_list.append(feat_df)
+        
         for raw_name, df in df_iterable:
             df = prepare_df(df, self.psm_df, column=self.column)
```
mschwoer, line 142: does this work? `self.column` is a list now ..

alphadia/outputtransform.py
```diff
@@ -125,45 +132,42 @@ def accumulate_frag_df(
...
-            quality_df = quality_df.merge(
-                df[["ion", "correlation", "precursor_idx"]],
-                on=["ion", "precursor_idx"],
-                how="outer",
-            )
-            quality_df.rename(columns={"correlation": raw_name}, inplace=True)
+        for idx in range(len(df_list)):
+            df_list[idx] = self._data_prep(df_list[idx])
+            
+        return df_list
```
mschwoer, line 155: `return [ self._data_prep(df) for df in df_list]`

alphadia/outputtransform.py
```diff
@@ -794,7 +798,9 @@ def build_lfq_tables(
         # as we want to retain decoys in the output we are only removing them for lfq
         qb = QuantBuilder(psm_df[psm_df["decoy"] == 0])
 
-        intensity_df, quality_df = qb.accumulate_frag_df_from_folders(folder_list)
+        dfs = qb.accumulate_frag_df_from_folders(folder_list)
+        intensity_df = dfs[0].copy()
```
mschwoer, line 802: the output of `accumulate_frag_df()` depends on the `column` parameter, right? 
hard-wiring this here to `intensity_df` & `quality_df` is very fragile ..

one solution could be that `accumulate_frag_df` returns a dictionary, mapping `columns` to the respective `df`
and then do
`intensity_df = dfs["intensity"]`

alphadia/outputtransform.py
```diff
@@ -794,7 +798,9 @@ def build_lfq_tables(
...
+        quality_df = dfs[2].copy()
```
mschwoer, line 803: why do we calculate a lot more now, if we then use only the first and third?

alphadia/outputtransform.py
```diff
@@ -296,7 +300,7 @@ def prepare_df(df, psm_df, column="intensity"):
         df["charge"].values,
         df["loss_type"].values,
     )
-    return df[["precursor_idx", "ion", column, "correlation"]]
+    return df[["precursor_idx", "ion"] + column ]
```
mschwoer, line 303: please change `column="intensity"` -> `columns=["intensity"])`

mschwoer (changes requested):
needs some work to make it more robust :-)

alphadia/outputtransform.py
vuductung: 👍

alphadia/outputtransform.py
vuductung: 👍

alphadia/outputtransform.py
vuductung: yes, since we're looping through the list of columns names (self.column in this case). The changes are mostly converting the data preparation from single handling `intensity_df` and `quality_df` to handling all "feature" dataframes.

alphadia/outputtransform.py
vuductung: more love added! :)

alphadia/outputtransform.py
vuductung: 👍

alphadia/outputtransform.py
vuductung: For the directLFQ quantification only intensity and correlation (first and third) features are needed. For future selectLFQ implementation all features are needed. I've made the changes for my personal usage, but I guess for the alphaDIA users, these features are not really used. Open for suggestions how we proceed from here.

alphadia/outputtransform.py
```diff
@@ -296,7 +306,7 @@ def prepare_df(df, psm_df, column="intensity"):
         df["charge"].values,
         df["loss_type"].values,
     )
-    return df[["precursor_idx", "ion", column, "correlation"]]
+    return df[["precursor_idx", "ion"] + columns]
```
mschwoer, line 309: `columns=["intensity"] -> columns=["intensity", "correlation"]`
otherwise, you're changing behaviour

alphadia/outputtransform.py
mschwoer: so this is required only once selectLFQ is integrated? then maybe change the default to `columns=["intensity", "correlation"]` for now and call the method with the extended `columns` argument in your own version?

mschwoer (commented):
fine from a code perspective, let's get @GeorgWa 's feedback also

alphadia/outputtransform.py
vuductung: yea you're totally right. I'll change the default `columns=["intensity", "correlation"]` and will call the other features in my own version.

alphadia/outputtransform.py
```diff
@@ -109,11 +110,8 @@ def accumulate_frag_df(
 
         Returns
         -------
-        intensity_df: pd.DataFrame
-            Dataframe with the intensity data containing the columns precursor_idx, ion, raw_name1, raw_name2, ...
-
-        quality_df: pd.DataFrame
-            Dataframe with the quality data containing the columns precursor_idx, ion, raw_name1, raw_name2, ...
+        dict
+            Dictionary of (column, df) tuples, where df is a dataframe with the columns precursor_idx, ion, raw_name1, raw_name2, ...
```
GeorgWa, line 114: Should we add that each key in the dict is an element in columns?

alphadia/outputtransform.py
```diff
@@ -794,7 +798,9 @@ def build_lfq_tables(
         # as we want to retain decoys in the output we are only removing them for lfq
         qb = QuantBuilder(psm_df[psm_df["decoy"] == 0])
 
-        intensity_df, quality_df = qb.accumulate_frag_df_from_folders(folder_list)
+        dfs = qb.accumulate_frag_df_from_folders(folder_list)
+        intensity_df = dfs["intensity"].copy()
```
GeorgWa, line 802: Do we need to call copy here? this is quite expensive I think and might not be necessary.

alphadia/outputtransform.py
```diff
@@ -849,7 +855,6 @@ class LFQOutputConfig:
 
             lfq_df = qb.lfq(
                 group_intensity_df,
-                quality_df,
```
GeorgWa, line 852: why are we dropping the quality df here?

GeorgWa (approved):
Can you add a minimal test case for the quant builder?
Looks good otherwise :) 

Do you think we can go one step further and move this to the Accumulator pattern which we are already using for the transfer learning? See here: https://github.com/MannLabs/alphadia/blob/main/alphadia/outputaccumulator.py

This would work out of memory and scale much better.

PR #501, opened 2025-03-13
move existing frozen_config files

mschwoer:
same mechanism as for log files: move existing `frozen_config.yaml` files


GeorgWa (approved):
I trust you that this makes it nicer from a user perspective!
LGTM ;)

mschwoer:
if we flood the directory with logs, flooding them additionally with configs is just as fine ;-)

PR #504, opened 2025-03-18
Fix multistep config

mschwoer:
fix wrong defaults for multistep search


GeorgWa (approved):
LGTM
I'm honestly a bit impressed. this was a bug which was really an edge case resulting from two separate code issues and we nailed it down immediately.

mschwoer:
image

PR #508, opened 2025-03-19
apply defaults on load

GeorgWa:
Fix the GUI where the default values were only loaded for parameters shown with the `ParameterInput` component.

This moves the initialization from defaults to a loading fuction.

### DIFF
--- a/gui/src/main/modules/workflows.js
+++ b/gui/src/main/modules/workflows.js
@@ -13,6 +13,7 @@ function discoverWorkflows(window){
             const workflow = JSON.parse(fs.readFileSync(path.join(directory, file)))
             workflow.name = path.basename(file, '.json')
             validateWorkflow(workflow)
+            initializeWorkflow(workflow)
             return [...acc, workflow]
         } catch (err) {
             dialog.showMessageBox(window, {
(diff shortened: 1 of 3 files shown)


gui/src/main/modules/workflows.js
```diff
@@ -74,6 +75,33 @@ function validateWorkflow(workflow) {
 
 }
 
+function initializeWorkflow(workflow) {
+    // Process each config section
+    workflow.config.forEach((config) => {
+        // Process regular parameters
+        if (config.parameters) {
```
mschwoer, line 82: (Bernie Meme) I am once again asking: do we need this distinction between `parameters` and `parameters_advanced`?

gui/src/main/modules/workflows.js
```diff
@@ -13,6 +13,7 @@ function discoverWorkflows(window){
             const workflow = JSON.parse(fs.readFileSync(path.join(directory, file)))
             workflow.name = path.basename(file, '.json')
             validateWorkflow(workflow)
+            initializeWorkflow(workflow)
```
mschwoer, line 16: this reads a bit wrong in terms of semantics ..
`validateWorkflowFile`?

mschwoer (approved):
LGTM .. adding one point to the "pain bucket" of "have a template for config"

jalew188:
Instead of the GUI, can we also provide a CLI, such as `alphadia default-config \path\to\config.file`, to generate a default config for users?

GeorgWa:
Yes @jalew188 . This should already be possible, its more abouit how the config is set in the GUI.

PR #516, opened 2025-03-27
Transferlearning on additional frag types

mo-sameh:
Using the adaptive fragment types introduced in https://github.com/MannLabs/alphapeptdeep/pull/228:
- Transfer learning now can be efficiently done with more fragment types than the default ones used in the pre-trained models.


alphadia/transferlearning/train.py
```diff
@@ -599,14 +610,17 @@ def finetune_ms2(
...
         # Test the model before training
-        self._test_ms2(
-            -1,
-            0,
-            reordered_val_psm_df,
-            reordered_val_intensity_df,
-            test_metric_manager,
-            data_split="validation",
-        )
+        if set(self.charged_frag_types).issubset(
```
GeorgWa, line 613: Do we need this check here? If yes, should there be an error message if this is not true?

GeorgWa (approved):
LGTM

PR #517, opened 2025-03-31
Display default value in parameter description

mschwoer:
This was lost somehow here https://github.com/MannLabs/alphadia/pull/479/files#diff-0fdfd7456a5b8c0ae939ca8329aee2cc35541ddfc4abd49103e518f5e228a1cdR239-R249

(main purpose was to make creation of guides easiert ;-))

### DIFF
--- a/gui/src/renderer/components/ParameterInput.js
+++ b/gui/src/renderer/components/ParameterInput.js
@@ -210,7 +210,7 @@ const ParameterInput = ({
 
     // make Grid which takes 100% of the height
     // The last row should grow to fill the remaining space
-    let default_text = parameter.type === "boolean" ? (parameter.default ? "true" : "false") : parameter.default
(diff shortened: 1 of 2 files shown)


gui/src/renderer/components/ParameterInput.js
```diff
@@ -223,15 +224,15 @@ const ParameterInput = ({
                 <Stack spacing={0.5}>
                     <Typography sx={{ fontWeight: 'bold' }}>{parameter.name}</Typography>
                     <Typography sx={{ fontFamily: 'monospace' }}>{`[${parameter_group_id}.${parameter.id}]`}</Typography>
```
GeorgWa, line 226: What do you think about moving this into the header? it's easy to miss at the end of the tooltip

GeorgWa (approved):
LGTM

PR #520, opened 2025-04-05
Don't perform FASTA reannotation in MBR step

GeorgWa:
There is a bug/misconfiguration where the second step in Library + MBR search is not using the protein inference from the first step.

This is due to the "AnnotateFasta" step being always applied as soon as a Fasta is provided.

Library search:
```
1 day, 2:33:53.742975 PROGRESS: ================ Protein FDR =================
1 day, 2:33:53.743260 PROGRESS: Unique protein groups in output
1 day, 2:33:53.743325 PROGRESS:   1% protein FDR: 10,016
1 day, 2:33:53.743354 PROGRESS: 
1 day, 2:33:53.743379 PROGRESS: Unique precursor in output
1 day, 2:33:53.743402 PROGRESS:   1% protein FDR: 138,918
1 day, 2:33:53.743425 PROGRESS: ================================================
```

Loading the library in the beginning of MBR:
```
1 day, 2:36:33.336888 INFO: ============ Library Stats ============
1 day, 2:36:33.336928 INFO: Number of precursors: 277,821
1 day, 2:36:33.393454 INFO: 	thereof targets:138,918
1 day, 2:36:33.393681 INFO: 	thereof decoys: 138,903
1 day, 2:36:33.396595 INFO: Number of elution groups: 138,918
1 day, 2:36:33.396716 INFO: 	average size: 2.00
1 day, 2:36:33.405208 INFO: Number of proteins: 10,442
1 day, 2:36:33.406364 INFO: Number of channels: 1 ([0])
1 day, 2:36:33.406480 INFO: Isotopes Distribution for 4 isotopes
1 day, 2:36:33.406513 INFO: =======================================
```

The solution is to drop fasta annotation in the MBR step.

### DIFF
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -269,7 +269,6 @@ multiplexing:
 
 fdr:
   fdr: 0.01
-  reannotate: False
   group_level: 'proteins'
   inference_strategy: "heuristic"
 
--- a/alphadia/constants/default.yaml
+++ b/alphadia/constants/default.yaml
@@ -269,6 +269,7 @@ multiplexing:
 
 fdr:
   fdr: 0.01
+  reannotate: False
   group_level: 'proteins'
   inference_strategy: "heuristic"
 
--- a/alphadia/constants/multistep.yaml
+++ b/alphadia/constants/multistep.yaml
@@ -34,6 +34,7 @@ library:
     enabled: False
 
 mbr:
+  fasta_paths: []
   fdr:
     inference_strategy: library
   search:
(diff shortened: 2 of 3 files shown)


mschwoer (approved):
lgtm.. although the "the "AnnotateFasta" step being always applied as soon as a Fasta is provided" ist quite an implicit behaviour that set this trap for the innocent developer of this feature

GeorgWa:
> lgtm.. although the "the "AnnotateFasta" step being always applied as soon as a Fasta is provided" ist quite an implicit behaviour that set this trap for the innocent developer of this feature

I thought the same and actually provided a first solution with a config key. But then I realized, that the only reason to provide a fasta is to perform 1) library prediction or 2) reannotation.

So there is no use-case in providing a fasta and not performing reannotation. Would be interested to hear som user perspective.

mschwoer:
maybe add a comment to the multistep config explaining why setting this is important

PR #521, opened 2025-04-09
Improve config handling

mschwoer:
Some improvements that made my life easier while testing different parameters.

### DIFF
--- a/alphadia/search_step.py
+++ b/alphadia/search_step.py
@@ -136,7 +136,7 @@ class SearchStep:
         default_config_path = os.path.join(
             os.path.dirname(__file__), "constants", "default.yaml"
         )
-        logger.info(f"loading config from {default_config_path}")
+        logger.info(f"loading default config from {default_config_path}")
         config = Config()
         config.from_yaml(default_config_path)
         return config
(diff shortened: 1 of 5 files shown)


alphadia/search_step.py
```diff
@@ -136,7 +136,7 @@ def _load_default_config():
         default_config_path = os.path.join(
             os.path.dirname(__file__), "constants", "default.yaml"
         )
-        logger.info(f"loading config from {default_config_path}")
+        logger.info(f"loading default config from {default_config_path}")
```
GeorgWa, line 139: I'm wondering, this is the default config. should we actually list the path at all as it's ot supposed to be changed (This could make it look like it).

docs/developer_guide.md
```diff
@@ -23,8 +23,8 @@ A good start for debugging is this notebook: `nvs/debug/debug_lvl1.ipynb`
  - "module": `alphadia.cli`
  - "script parameters": `--config /abs/path/to/tests/e2e_tests/basic/config.yaml`
  - "working directory": `/abs/path/to/tests/e2e_tests`
-2. Uncomment the lines following the `uncomment for debugging` comment in `alphadia/cli.py`.
-3. Run the configuration.
+ - "Environment variables": `RUN_MAIN=1`
```
GeorgWa, line 26: What does this do? Is this consumed by alphaDIA?

GeorgWa (approved):
LGTM

docs/developer_guide.md
mschwoer: yes, here https://github.com/MannLabs/alphadia/pull/521/files#diff-e438c20d5cb43d2b92918217e0b7ef0e29e3935156a0debac075b6cb700d9258R261

PR #522, opened 2025-04-09
Improve method section for optimization

GeorgWa:
Updated the method section in the paper and also updated it in the docs.


mschwoer (commented):
LGTM

machth:
24 1. Search is performed batch-wise, starting with the first 8000 precursors => maybe specify more: are those the top 8000 precursors, first 8000 precursors from the library , random first 8000 precursors or first 8000 found precursors?

Otherwise LGTM!

PR #539, opened 2025-04-11
Speed up -v and -h

mschwoer:
Speeding up -v and -h, which take considerably longer since the dependencies have been upgraded (due to module import)

### DIFF
--- a/alphadia/cli.py
+++ b/alphadia/cli.py
@@ -14,15 +14,11 @@ import sys
 
 import yaml
 
+from alphadia import __version__  # noqa: E402
+from alphadia.constants.keys import ConfigKeys
+
 logging.getLogger("matplotlib.font_manager").setLevel(logging.ERROR)
 import matplotlib  # noqa: E402
-
-import alphadia
-from alphadia import utils
-from alphadia.constants.keys import ConfigKeys
-from alphadia.exceptions import CustomError
-from alphadia.search_plan import SearchPlan
-from alphadia.workflow import reporting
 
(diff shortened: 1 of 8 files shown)


alphadia/cli.py
```diff
@@ -112,6 +116,30 @@
 )
 
 
+def _recursive_update(
+    full_dict: dict, update_dict: dict
+):  # TODO merge with Config._update
```
GeorgWa, line 121: Can't we use the config update logic for this?

alphadia/cli.py
```diff
@@ -189,28 +217,41 @@ def _get_raw_path_list_from_args_and_config(
...
+        return
+
+    # load modules only here to speed up -v and -h commands
+    from alphadia.exceptions import CustomError
+    from alphadia.search_plan import SearchPlan
+    from alphadia.workflow import reporting
+
+    if args.check:
+        print(
+            f"{__version__}"
```
GeorgWa, line 238: I would suggest to return a json here:
`{"version: "1.10.3", "import_success": True}`

Can we wrap the imports in a try except so that we could issue on failure:
`{"version: "1.10.3", "import_success": False}`

GeorgWa (approved):
Do we know why the imports take longer since the upgrade of the dependencies?
LGTM

mschwoer:
> Do we know why the imports take longer since the upgrade of the dependencies? LGTM

random pause method says numba, but I did not dig into it

GeorgWa:
Let's merge it like this :)

PR #544, opened 2025-04-23
Fdr memopt

mschwoer:
Changing the way the score cutoff is computed.


GeorgWa (approved):
Funny how little code this is compared to the benchmarking :D

PR #551, opened 2025-05-21
Filter fragments df

mschwoer:
- avoid running into "fragment_mz_tolerance must be less than 200" error by <del>renaming it to "fragment_mz_tolerance must be less than or equal 200" </del> filtering it out before hand
- introduced `settings.py` to collect all magic numbers


alphadia/workflow/peptidecentric.py
```diff
@@ -708,34 +709,40 @@ def filter_dfs(self, precursor_df, fragments_df):
...
+            precursor_idx_mask & mass_error_mask
+        ].sort_values(by="correlation", ascending=False)
 
-        fragments_df_filtered = fragments_df_filtered.sort_values(
-            by="correlation", ascending=False
-        )
         # Determine the number of fragments to keep
-        min_fragments, max_fragments = (
-            500,
+        min_fragments = 500  # TODO remove min_fragments as it seems to have no effect
```
GeorgWa, line 730: yes, can indeed be removed

GeorgWa (approved):
Looks good

PR #552, opened 2025-05-21
Invalid dia data error

mschwoer:


alphadia/data/alpharaw_wrapper.py
```diff
@@ -242,9 +74,28 @@ def calculate_valid_scans(quad_slices: np.ndarray, cycle: np.ndarray):
     return np.array(precursor_idx_list)
 
 
-class AlphaRaw(alpharaw_thermo.MSData_Base):
+class AlphaRaw(MSData_Base):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        self.sample_name = None
```
GeorgWa, line 81: why are we adding the variables here?

I would think some of them could overwrite values defined in `MSData_Base.init()` and most are only part of `AlphaRawJIT`

GeorgWa (approved):
LGTM

PR #554, opened 2025-05-22
Remove neptune from code

mschwoer:
Remove all references to neptune from the production code


GeorgWa (approved):
Long awaited :D
As far as I remember we already removed it from the requirements?

PR #555, opened 2025-05-22
Re enable plotting

mschwoer:
Re-enable the plotting functionality, making it switchable by default.

I re-installed the calibration plots, does this make sense @GeorgWa ?

### DIFF
--- a/alphadia/workflow/base.py
+++ b/alphadia/workflow/base.py
@@ -4,9 +4,9 @@ import os
 
 # alpha family imports
 from alphabase.spectral_library.base import SpecLibBase
-from constants.settings import FIGURES_FOLDER_NAME
 
 from alphadia.constants.keys import ConfigKeys
+from alphadia.constants.settings import FIGURES_FOLDER_NAME
 
 # alphadia imports
 from alphadia.data import alpharaw_wrapper, bruker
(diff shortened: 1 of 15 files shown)


alphadia/workflow/base.py
```diff
@@ -56,6 +56,12 @@ def __init__(
         self._quant_path: str = quant_path or os.path.join(
             config[ConfigKeys.OUTPUT_DIRECTORY], QUANT_FOLDER_NAME
         )
+        self.figure_path: str = (
```
GeorgWa, line 59: This is the part of the quant folder. Do I understand this correctly?
Why are we not using `self._quant_path` here?

GeorgWa: OOhhh lol, I missunderstood what quant path is. Should we add @properties with a docstring for the two of them?

tests/unit_tests/test_outputaccumulator.py
```diff
@@ -37,6 +37,7 @@ def prepare_input_data():
     config = {
         "general": {
             "thread_count": 8,
+            "save_figures": True,
```
GeorgWa, line 40: Good default!

GeorgWa (approved):
LGTM

alphadia/workflow/base.py
mschwoer: I would rather remove some properties :-)

mschwoer: hm, isn't adding a (unused) propertyquite some overhead for documenting an instance variable? cf. https://github.com/MannLabs/alphadia/pull/558

PR #558, opened 2025-06-02
Refactor workflow paths

mschwoer:
simplify the paths in the workflow

### DIFF
--- a/alphadia/fdr.py
+++ b/alphadia/fdr.py
@@ -223,7 +223,7 @@ def fdr_to_q_values(fdr_values: np.ndarray):
     return q_vals
 
 
-def q_values(
+def q_values(  # TODO unused
     scores: np.ndarray,
     decoy_labels: np.ndarray,
     # score_column : str = 'proba',
@@ -301,7 +301,7 @@ def get_q_values(
     target_values = 1 - _df[decoy_column].values
     decoy_cumsum = np.cumsum(_df[decoy_column].values)
     target_cumsum = np.cumsum(target_values)
-    fdr_values = decoy_cumsum / target_cumsum
+    fdr_values = decoy_cumsum / target_cumsum  # here
     _df[qval_column] = fdr_to_q_values(fdr_values)
     return _df
 
--- a/alphadia/workflow/base.py
+++ b/alphadia/workflow/base.py
@@ -53,16 +53,20 @@ class WorkflowBase:
 
         """
         self._instance_name: str = instance_name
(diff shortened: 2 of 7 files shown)


alphadia/fdr.py
```diff
@@ -223,7 +223,7 @@ def fdr_to_q_values(fdr_values: np.ndarray):
     return q_vals
 
 
-def q_values(
```
GeorgWa, line 226: was this sneaked in? :D

alphadia/workflow/base.py
```diff
@@ -141,20 +145,10 @@ def load(
...
-        return self._instance_name
-
-    @property
-    def quant_path(self) -> str:
-        """Path where the workflow folder will be created"""
-        return self._quant_path
-
     @property
     def path(self) -> str:
         """Path to the workflow folder"""
```
GeorgWa, line 150: Maybe add an example `first_search/quant/raw_file_xyz.raw`

alphadia/workflow/base.py
```diff
@@ -74,7 +78,7 @@ def __init__(
         self._optimization_manager: manager.OptimizationManager | None = None
```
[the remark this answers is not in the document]
GeorgWa: I agree with Mr. AI 😄 , do we want to remove `self._quant_path` ?

alphadia/workflow/base.py
```diff
@@ -141,20 +145,10 @@ def load(
 
         self.reporter.log_event("section_stop", {})
 
-    @property
-    def instance_name(self) -> str:
-        """Name for the particular workflow instance. this will usually be the name of the raw file"""
-        return self._instance_name
-
-    @property
-    def quant_path(self) -> str:
```
GeorgWa, line 150: what is the reason that these two were removed but .path not? just to slim down the interface? Would be fine by me, just curious.

GeorgWa (approved):
LGTM

alphadia/fdr.py
mschwoer: the comment? yes .. the method? idk :-)
We should remove all unused code or at least comment on it why it should stay

alphadia/workflow/base.py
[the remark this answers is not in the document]
mschwoer: iirc, it's not used anywhere except in the constructor .. and as `path` is derived from `quant_path`, the latter does not need to be created explicitly .. or am I missing smth?

PR #560, opened 2025-06-03
Split outputtransform

mschwoer:
split outputtransform

only moved code


GeorgWa (approved):
I think this rebuild makes lot's of sense. I like the result.

PR #559, opened 2025-06-03
Split features

mschwoer:
Splitting the `features.py` module and move it to `plexscoring`

Only moved code

PR #567, opened 2025-06-10
update documentation configuration and add autosummary for modules

mschwoer:
fix the docs that I broke in https://github.com/MannLabs/alphadia/pull/559 and https://github.com/MannLabs/alphadia/pull/560


GeorgWa (approved):
LGTm

PR #573, opened 2025-06-25
Minor refactorings

mschwoer:
- don't store whole config in Optimization
- mark private methods as private


GeorgWa (approved):
I trust you that all of the renamings were sensible, looks good!

PR #576, opened 2025-06-26
split_fdr_I

mschwoer:
Rename fdrx to indicate it should not be used, move out those methods that are used to `fdr_` (temporary name)
Delete some unused methods & Classifiers

Up next: split `fdr.py` and `fdrexperimental.py` into modules of `fdr_` (which will then be renamed to `fdr`)


GeorgWa (approved):
Fits to what we discussed :D

PR #588, opened 2025-06-27
Split validate

mschwoer:
- split validation into a package
- make schema validation an explicit all (cf. nest PR)

### DIFF
--- a/alphadia/peakgroup/search.py
+++ b/alphadia/peakgroup/search.py
@@ -16,7 +16,12 @@ from alphadia.peakgroup.utils import (
     find_peaks_2d,
 )
 from alphadia.utils import USE_NUMBA_CACHING
-from alphadia.validation import validate
+from alphadia.validation.schemas import (
+    fragments_flat as validate_fragments_flat,
+)
+from alphadia.validation.schemas import (
+    precursors_flat as validate_precursors_flat,
+)
 
 logger = logging.getLogger()
 
@@ -938,7 +943,7 @@ class HybridCandidateSelection:
             )
 
         # validate dataframe schema and prepare jitclass compatible dtypes
-        validate.fragments_flat(self.fragments_flat)
+        validate_fragments_flat(self.fragments_flat)
 
         return fragments.FragmentContainer(
             self.fragments_flat["mz_library"].values,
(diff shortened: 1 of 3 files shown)


alphadia/validation/schemas.py
```diff
@@ -317,37 +132,17 @@ def candidates_df(df: pd.DataFrame, logging: bool = True):
...
     """
     check_critical_values(df)
-    candidates_schema(df, logging=logging)
+    candidates_schema.validate(df, logging=logging)
 
 
 candidates_df.__doc__ += candidates_schema.docstring()
 
 
-def check_critical_values(input_df):
```
GeorgWa, line 332: I know it's not validation, but this was quite useful to figure out if something is wrong

alphadia/validation/base.py
```diff
@@ -0,0 +1,196 @@
...
+            List of Property objects
+
+        """
+        self.name = name
+        self.schema = properties
+        for property in self.schema:
+            if not isinstance(property, Property):
+                raise ValueError("Schema must contain only Property objects")
+
+    def validate(self, df: pd.DataFrame, logging: bool = True) -> None:
```
GeorgWa, line 126: this feels a bit sneaked in ;D
WOuld you generally recommend against using `__call__` ?

GeorgWa (commented):
I would like to keep the nan functionality, maybe in a separate function

alphadia/validation/base.py
mschwoer: In most cases, I feel `__call__` makes things harder to read .. if you have something like 
```
validator = Validator(myparam=3)
validator.validate(my_data)
```
then you have a little reduncancy (`validator.validate`) but it makes sense in terms of semantics ..

in contrast
```
validator = Validator(myparam=3)
validator(my_data)
```
feels off to me (and to LLMs?)



Edit: no
Here are 10 independent judging rounds:
Round 1: __call__ - More Pythonic and follows the principle of least surprise. Objects that do one main thing should be callable.
Round 2: .validate() - Explicit is better than implicit. The method name clearly states the intent, making code more self-documenting.
Round 3: __call__ - Cleaner syntax reduces visual noise. When you're chaining operations or using it frequently, the brevity matters.
Round 4: .validate() - Better for team codebases. New developers can immediately understand what the method does without knowing the class internals.
Round 5: __call__ - Follows functional programming patterns. Makes the validator feel like a configured function rather than a stateful object.
Round 6: .validate() - More maintainable. If you later need multiple methods like .validate_async() or .validate_partial(), having explicit names is clearer.
Round 7: __call__ - Consistent with Python's ecosystem. Libraries like scikit-learn, PyTorch, and many others use callable objects extensively.
Round 8: .validate() - Better IDE support. Autocomplete and type hints work more reliably with explicit method names.
Round 9: __call__ - Reduces cognitive load. You think "validate this data" not "call the validate method on this validator with this data."
Round 10: .validate() - Safer for debugging. Stack traces and error messages are clearer when they show explicit method names.
Final tally: __call__ wins 5-5... it's a tie! Both approaches have strong merits depending on your specific context and team preferences.

PR #591, opened 2025-06-30
Refactor numba

mschwoer:
- move the contents of the `numba` module
- delete some obsolete/unused functions


tests/unit_tests/test_search_utils.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+
+            assert limits[0] <= limits[1]
+            assert limits[0] >= 0
+
+            assert limits[0] <= center
+            assert limits[1] >= center
+
+
+@skip("raises NumbaContextOnly")
```
GeorgWa, line 37: I can't find the function in this diff. Where is it from?

alphadia/utilities/fragment_container.py
```diff
@@ -0,0 +1,130 @@
...
+        self.mz = self.mz[mz_order]
+        self.intensity = self.intensity[mz_order]
+        self.type = self.type[mz_order]
+        self.loss_type = self.loss_type[mz_order]
+        self.charge = self.charge[mz_order]
+        self.number = self.number[mz_order]
+        self.position = self.position[mz_order]
+        self.cardinality = self.cardinality[mz_order]
+
+    def filter_by_cardinality(self, max_cardinality: nb.uint8):
```
GeorgWa, line 66: we can check which are actually used and remove the remaining once

tests/unit_tests/test_search_utils.py
GeorgWa: found it. Why is it failing now but wasn't failing before?

alphadia/utilities/fragment_container.py
```diff
@@ -0,0 +1,130 @@
+import numba as nb
```
GeorgWa, line 1: I think the naming is a bit strange to have both `utils.py` and `utilities/*`.
Do you think this could be changed?

GeorgWa (approved):
Have a look at the folder structure. Other than this it's good.

PR #593, opened 2025-06-30
move extraction_handler.py out of peptidecentric.py

mschwoer:
A suggestion on how to split `peptidecentric.py`..
the state would still be stored in the `workflow` object, but the operations are done (directly on that state) by dedicated classes ..

Not sure if this is the way to go though


mschwoer:
see https://github.com/MannLabs/alphadia/pull/602

PR #602, opened 2025-07-03
Split peptidecentric i

mschwoer:
The handsome brother of https://github.com/MannLabs/alphadia/pull/593

### DIFF
--- a/alphadia/peakgroup/config_df.py
+++ b/alphadia/peakgroup/config_df.py
@@ -124,7 +124,9 @@ class HybridCandidateConfigJIT:
         self.feature_weight = feature_weight
 
 
-class HybridCandidateConfig(JITConfig):
+class HybridCandidateConfig(
+    JITConfig
+):  # TODO rename to HybridCandidateScoringHyperparameters
     jit_container = HybridCandidateConfigJIT
 
     def __init__(self):
--- a/alphadia/workflow/peptidecentric/extraction_handler.py
+++ b/alphadia/workflow/peptidecentric/extraction_handler.py
@@ -52,9 +52,9 @@ class ExtractionHandler:
         )
 
         scoring_config = HybridCandidateConfig()
-        scoring_config.update(self._config["selection_config"])
         scoring_config.update(
             {
+                **self._config["selection_config"],
                 "top_k_fragments": self._config["search"]["top_k_fragments"],
                 "rt_tolerance": self._optimization_manager.rt_error,
                 "mobility_tolerance": self._optimization_manager.mobility_error,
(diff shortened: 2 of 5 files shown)


alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -180,6 +175,17 @@ def load(
         self._init_fdr_manager()
         self._init_spectral_library()
 
+        self._extraction_handler = ExtractionHandler(
+            self.config,
+            self.optimization_manager,
+            self.reporter,
+            self.spectral_library,
+            rt_column=self._get_rt_column(),
```
mschwoer, line 183: this is a bug. the column names change during the processing .. will be changed in another PR

mschwoer: https://github.com/MannLabs/alphadia/pull/611

alphadia/peakgroup/config_df.py
```diff
@@ -124,7 +124,9 @@ def __init__(
         self.feature_weight = feature_weight
 
 
-class HybridCandidateConfig(JITConfig):
+class HybridCandidateConfig(
+    JITConfig
+):  # TODO rename to HybridCandidateScoringHyperparameters
```
GeorgWa, line 129: I think we can simplify the name and drop the `Hybrid` to `CandidateScoring` and

alphadia/workflow/peptidecentric/extraction_handler.py
```diff
@@ -0,0 +1,140 @@
...
+from alphadia.peakgroup.config_df import HybridCandidateConfig
+from alphadia.plexscoring.config import CandidateConfig
+from alphadia.plexscoring.plexscoring import CandidateScoring
+from alphadia.reporting.reporting import Pipeline
+from alphadia.workflow.config import Config
+from alphadia.workflow.managers.optimization_manager import OptimizationManager
+
+
+class ExtractionHandler:
+    """Manages precursor and fragment extraction operations."""
```
GeorgWa, line 17: I would make the Handler pattern a bit more clear in the docstring. Maybe that this is an operation as part of a search workflow for a single file and that it should be used as an operator pattern. SO the object has no state but is initialized when it's needed.

alphadia/workflow/peptidecentric/extraction_handler.py
```diff
@@ -0,0 +1,140 @@
...
+        reporter: Pipeline,
+        spectral_library: SpecLibBase,
+        *,
+        rt_column: str,
+        mobility_column: str,
+        precursor_mz_column: str,
+        fragment_mz_column: str,
+    ):
+        self._config: Config = config
+        # TODO think about passing only what is needed, e.g. rt_error, ms1_error, etc.
```
GeorgWa, line 32: I'm not sure if this is a good pattern. I would rather encapsulate the state in context objects.

GeorgWa (approved):
Looks good

PR #604, opened 2025-07-03
move out requantification_handler.pyi

mschwoer:
move out requantification handler

@GeorgWa I guess both of these usecases are not covered by e2e tests?

### DIFF
--- a/alphadia/workflow/peptidecentric/peptidecentric.py
+++ b/alphadia/workflow/peptidecentric/peptidecentric.py
@@ -921,7 +921,6 @@ class PeptideCentricWorkflow(base.WorkflowBase):
         if not self._requantification_handler:
             self._requantification_handler = RequantificationHandler(
                 self.config,
-                self.optimization_manager,
                 self.calibration_manager,
                 self.fdr_manager,
                 self.reporter,
--- a/alphadia/workflow/peptidecentric/peptidecentric.py
+++ b/alphadia/workflow/peptidecentric/peptidecentric.py
@@ -154,7 +154,7 @@ class PeptideCentricWorkflow(base.WorkflowBase):
 
         self._extraction_handler: ExtractionHandler | None = None
         self._recalibration_handler: RecalibrationHandler | None = None
-        self.requantification_handler: RequantificationHandler | None = None
+        self._requantification_handler: RequantificationHandler | None = None
 
     def load(
         self,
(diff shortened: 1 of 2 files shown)


alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -920,110 +916,29 @@ def log_precursor_df(self, precursor_df):
...
+    def requantify(self, psm_df: pd.DataFrame) -> pd.DataFrame:
+        """TODO.
 
-        target_channels = [
-            int(c) for c in self.config["multiplexing"]["target_channels"].split(",")
-        ]
-        reference_channel = self.config["multiplexing"]["reference_channel"]
+        Delegates to RequantificationHandler.requantify(), see docstring there for more details.
+        """
+        self._lazy_init_requantification_handler()
```
GeorgWa, line 939: can't we move this code here? I'm not sure how much we are winning by isolating it into a function

alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -159,6 +154,7 @@ def __init__(
 
         self._extraction_handler: ExtractionHandler | None = None
         self._recalibration_handler: RecalibrationHandler | None = None
+        self._requantification_handler: RequantificationHandler | None = None
```
GeorgWa, line 157: We are changing the semantics of the handlers here a bit.

In `PeptideCentric` there are certain high level steps which are actually called like a pipeline if needed:

`PeptideCentric.search_parameter_optimization`
`PeptideCentric.extraction`
`PeptideCentric.requantify`
`PeptideCentric.requantify_fragments`

The requantification handler is now very close to requantify. I think we should make this hierarchy more clear.

alphadia/workflow/peptidecentric/requantification_handler.py
```diff
@@ -0,0 +1,348 @@
...
+    candidate_features_to_candidates,
+    multiplex_candidates,
+)
+from alphadia.reporting.reporting import Pipeline
+from alphadia.workflow.config import Config
+from alphadia.workflow.managers.calibration_manager import CalibrationManager
+from alphadia.workflow.managers.fdr_manager import FDRManager
+
+
+class RequantificationHandler:
```
GeorgWa, line 21: I think we shouldn't put them in one handler. The naming is similar but it's actually two different things not really connected.

One is needed for transfer learning the other is needed for multiplexing.

GeorgWa:
> move out requantification handler
> 
> @GeorgWa I guess both of these usecases are not covered by e2e tests?

One use case is covered with transfer learning activated

alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -920,110 +916,29 @@ def log_precursor_df(self, precursor_df):
...
-            f"decoy channel: {decoy_channel}", verbosity="progress"
-        )
-
-        channels = list(
-            set(
-                original_channels
-                + [reference_channel]
-                + target_channels
-                + [decoy_channel]
+    def _lazy_init_requantification_handler(self):
```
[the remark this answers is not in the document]
mschwoer: no longer, for several reasons:
1) split up `requantification_handler`
2) initialize on the fly 
:-)

PR #606, opened 2025-07-03
Split peptidecentric iv

mschwoer:
moving out optimization. see individual commits for how this was approached..


alphadia/workflow/peptidecentric/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        precursor_001fdr = len(
+            precursor_df[
+                (precursor_df["qval"] < 0.001)
+                & (precursor_df["decoy"] == 0)
+                & (precursor_df["channel"] == channel)
+            ]
+        )
+        reporter.log_string(
+            f"Channel {channel:>3}:\t 0.05 FDR: {precursor_05fdr:>5,}; 0.01 FDR: {precursor_01fdr:>5,}; 0.001 FDR: {precursor_001fdr:>5,}",
```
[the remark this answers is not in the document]
GeorgWa: good idea!

alphadia/workflow/peptidecentric/optimization_handler.py
```diff
@@ -0,0 +1,445 @@
...
+        extraction_handler: ExtractionHandler,
+        recalibration_handler: RecalibrationHandler,
+        reporter: Pipeline,
+        spectral_library: SpecLibBase,
+        dia_data: AlphaRaw | TimsTOFTranspose,
+    ):
+        self.config = config
+        # TODO decouple optimizers from workflow:
+        #  there's an implicit coupling between this class and the Optimizers, which think this is a workflow object!
+        self.optimization_manager = optimization_manager
```
GeorgWa, line 39: I think here we really define the code pattern better. I would be very confused if there are both optimization handler and optimization manager :D

GeorgWa: I think making this an `OptimizationContext` would be the cleanest

alphadia/workflow/peptidecentric/utils.py
```diff
@@ -0,0 +1,117 @@
+import numpy as np
+import pandas as pd
+
+from alphadia.reporting.reporting import Pipeline
+from alphadia.workflow.config import Config
+from alphadia.workflow.managers.fdr_manager import FDRManager
+
+
+def fdr_correction(
```
GeorgWa, line 9: I think we could actually have an FDRHandler to do this. What happening here is the same as in the other handlers.

tests/unit_tests/workflow/test_peptidecentric.py
```diff
@@ -31,10 +32,21 @@ def test_filters_precursors_and_fragments_correctly(mock_config):
     )
     instance = PeptideCentricWorkflow("test_instance", mock_config)
     instance.reporter = MagicMock()
+    instance._optimization_handler = OptimizationHandler(
```
GeorgWa, line 35: what is happening here? :D

GeorgWa (approved):
LGTM for now.
But I think we have to sharpen the pattern a bit. I would also suggest to move the FDR from utils to a handler.

PR #607, opened 2025-07-03
Split peptidecentric v

mschwoer:
some more refactorings, see individual commits

Claude's take on the efforts of splitting peptidecentric so far: 
>  The structure shows good intent but suffers from tight coupling, complex dependencies, and some implementation issues that make it difficult to maintain and test.

for sure will add "shows good intent" to my passive-aggressive toolbox :-D


alphadia/workflow/peptidecentric/library_init.py
```diff
@@ -0,0 +1,128 @@
+import numpy as np
+from alphabase.spectral_library.base import SpecLibBase
+
+from alphadia.reporting.reporting import Pipeline
+from alphadia.workflow.config import Config
+
+
+def init_spectral_library(
```
GeorgWa, line 8: we could have an init handler to decouple it.. :D

GeorgWa (approved):

PR #609, opened 2025-07-04
Simplify init library

mschwoer:
as discussed:
- remove using `config["search_initial"]["initial_rt_tolerance"]`: this has never been adapted to `initial_rt_tolerance` being small now by default (3e0b7424) and therefore should have no effect (according to @GeorgWa )
- unused "mode" has been removed
- add option to config to tolerate removed keys in order to prevent breaking changes


GeorgWa (approved):
This makes my previous comment irrelevant

PR #610, opened 2025-07-04
Fix neptune in e2e

mschwoer:
- metrics calculation stopped working a while ago due to removal of `ms1_accuracy`
- changed file tracking to actual file upload @GeorgWa was there a reason not to due this in the first place?

before fix:
image

after:

image

image

image


GeorgWa (approved):
Looks good, no I dont think there was a reason :D

PR #611, opened 2025-07-04
split_peptidecentric_VII

mschwoer:
fix bug introduced in #602 (https://github.com/MannLabs/alphadia/pull/602/files#r2185648812)


mschwoer:
> I would actually suggest to merge this with the `optimization_manager` as it's just a lightweight wrapper around it. What do you think? can be done later though

there's a subtlety there: the `ColumnNameHandler` needs information from the `dia_data` object, the `optimization_manager` does not ..

PR #618, opened 2025-07-10
Refactor managers vi

mschwoer:
Make the implicit dependencies through workflow explicit

### DIFF
--- a/alphadia/workflow/peptidecentric/optimization_handler.py
+++ b/alphadia/workflow/peptidecentric/optimization_handler.py
@@ -49,8 +49,6 @@ class OptimizationHandler:
         figure_path: str | None = None,
     ):
         self.config = config
-        # TODO decouple optimizers from workflow:
-        #  there's an implicit coupling between this class and the Optimizers, which think this is a workflow object!
         self.optimization_manager = optimization_manager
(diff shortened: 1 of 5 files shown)


alphadia/workflow/peptidecentric/optimization_handler.py
```diff
@@ -60,7 +58,39 @@ def __init__(
...
+
+    def _init_optimiser(
+        self,
+        clazz: type[AutomaticOptimizer | TargetedOptimizer],
+        initial_parameter: float,
+        target_parameter: float | None = None,
+    ):
+        """Returns an instance of the specified optimizer class with the given initial and target parameters."""
+        if issubclass(clazz, TargetedOptimizer):
+            return clazz(
```
GeorgWa, line 73: "Clazz" can refer to a Java type used in programming or a German a cappella choir. In Java, "clazz" is a common variable name used to represent a class object, often when the keyword "class" is already in use. "CLAZZ" (with all caps) is also the name of a German choir specializing in a cappella arrangements of pop and jazz music.

GeorgWa: Learned something today

GeorgWa (approved):
LGTM

PR #620, opened 2025-07-11
Refactor data

mschwoer:
- move out the JIT classes, as they are very close to actual search
- renamings, docstrings


GeorgWa (approved):
LGTM

PR #622, opened 2025-07-14
Misc refactorings iii

mschwoer:
some more corrections & simplifications, see individual commits


alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -164,12 +164,12 @@ def search_parameter_optimization(self):
 
         self._save_managers()
 
-    def extraction(self):
         self.calibration_manager.predict(
             self.spectral_library.precursor_df, "precursor"
         )
-        self.calibration_manager.predict(self.spectral_library._fragment_df, "fragment")
+        self.calibration_manager.predict(self.spectral_library.fragment_df, "fragment")
```
mschwoer, line 170: note that I moved this up here, making "prepare speclib with optimization results" part of the the optimization ..

PR #628, opened 2025-07-18
More timings

mschwoer:
add timings for loading the raw file and the whole processing (to be able to identify other bottlenecks)

### DIFF
--- a/alphadia/workflow/peptidecentric/peptidecentric.py
+++ b/alphadia/workflow/peptidecentric/peptidecentric.py
@@ -24,6 +24,7 @@ from alphadia.workflow.peptidecentric.requantification_handler import (
 from alphadia.workflow.peptidecentric.utils import (
     feature_columns,
     log_precursor_df,
+    use_timing_manager,
 )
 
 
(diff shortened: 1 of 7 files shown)


alphadia/workflow/peptidecentric/peptidecentric.py
```diff
@@ -87,6 +91,12 @@ def __init__(
         )
         self._fdr_manager: FDRManager | None = None
 
+        self._timing_manager: TimingManager = TimingManager(
+            path=os.path.join(self.path, self.TIMING_MANAGER_PKL_NAME),
+            load_from_file=self.config["general"]["reuse_calibration"],
+        )
+
+    @use_timing_manager("load")
```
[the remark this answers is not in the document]
GeorgWa: good idea! Could be from @mschwoer

tests/unit_tests/workflow/test_peptidecentric.py
```diff
@@ -12,7 +12,7 @@
 @pytest.fixture
 def mock_config():
     return {
-        "general": {"save_figures": True},
+        "general": {"save_figures": True, "reuse_calibration": False},
```
GeorgWa, line 15: isn't this the default? why do we need to set this?

GeorgWa (approved):
LGTM

tests/unit_tests/workflow/test_peptidecentric.py
mschwoer: this is just required to set up the mock test data (which does not read the `default.yaml`)

PR #632, opened 2025-07-22
remove previously removed classifier from fdr performance test

mschwoer:
remove previously removed classifier from fdr performance test


GeorgWa (approved):
As mentioned, we can deprecate the fdr performance test for now. mit rebuild it in the future

PR #637, opened 2025-07-24
Misc feedback

mschwoer:
Addressing the feedback from previous PRs, and some small tweaks along the way. See individual commits.

### DIFF
--- a/alphadia/fragcomp/fragcomp.py
+++ b/alphadia/fragcomp/fragcomp.py
@@ -1,11 +1,13 @@
 """The fragment competition module contains functionality to maintain the exclusive assignment of signal to identifications."""
 
 import logging
+import warnings
 
 import numba as nb
 import numpy as np
 import pandas as pd
 from alphatims import utils as timsutils
+from pandas.errors import SettingWithCopyWarning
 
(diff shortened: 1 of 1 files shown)


alphadia/fragcomp/fragcomp.py
```diff
@@ -245,6 +245,8 @@ def __call__(
             The PSM dataframe with the valid column.
 
         """
+        psm_df = psm_df.copy()  # to avoid SettingWithCopyWarning
```
GeorgWa, line 248: this solves the symptom but is rather memory expensive. Could we try to solve this at the origin?
We might have a `psm_df = xyz_df[some_filter]` somewhere leaving `xyz_df` dangling.

GeorgWa (approved):
LGTM

alphadia/fragcomp/fragcomp.py
mschwoer: good point.. 
if we have this outside
`psm_df = psm_df[..].copy()` .. according to some research & experiments, garbage collection will remove the original dataframe immediately, seemingly entailing no extra memory requirements .. but not sure if there's still a spike? .. added a TODO

PR #638, opened 2025-07-24
add DiaData interface

mschwoer:
Add an explicit interface for DiaData


alphadia/raw_data/alpharaw_wrapper.py
```diff
@@ -10,12 +10,13 @@
...
 
 from alphadia.raw_data.dia_cycle import determine_dia_cycle
+from alphadia.raw_data.interface import DiaData
 from alphadia.raw_data.jitclasses.alpharaw_jit import AlphaRawJIT
 
 logger = logging.getLogger()
 
 
-class AlphaRaw(MSData_Base, ABC):
+class AlphaRaw(MSData_Base, DiaData, ABC):
```
GeorgWa, line 19: Image

GeorgWa, line 19: Why do we need the ABC here?

GeorgWa (approved):
LGTM, still not quite clear how I can test against this from Rust

PR #641, opened 2025-07-30
add first tests for quant_builder

mschwoer:
Add unit tests to prepare for an extension of functionality

### DIFF
--- a/tests/unit_tests/test_quant_builder.py
+++ b/tests/unit_tests/test_quant_builder.py
@@ -152,11 +152,11 @@ class TestQuantBuilderAccumulateFragDf:
     ):
         """Test that accumulate_frag_df returns correct dataframes for single run."""
         # given
-        builder = QuantBuilder(extended_psm_df, column="intensity")
+        builder = QuantBuilder(extended_psm_df, columns=["intensity", "correlation"])
         df_iterable = [("run1", simple_fragment_df)]
 
         # when
-        intensity_df, quality_df = builder.accumulate_frag_df(iter(df_iterable))
+        result_dict = builder.accumulate_frag_df(iter(df_iterable))
 
         # then
         expected_intensity_df = pd.DataFrame(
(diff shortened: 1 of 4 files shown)


tests/unit_tests/test_quant_builder.py
```diff
@@ -0,0 +1,577 @@
...
+            "intensity": [100, 200, 300, 400, 500],
+            "correlation": [0.8, 0.9, 0.7, 0.6, 0.5],
+        }
+    )
+
+
+class TestQuantBuilderAccumulateFragDf:
+    """Test cases for QuantBuilder.accumulate_frag_df() method."""
+
+    def test_accumulate_frag_df_single_run_should_return_intensity_and_quality_dfs(
```
mschwoer, line 150: @GeorgWa so far I did not like this nested structure with the extra `class`, but maybe it is better suited for LLMs?

GeorgWa: same here.. Maybe it's the way to go. It also allows to give a context for fixtures etc. which I like

tests/unit_tests/test_quant_builder.py
```diff
@@ -0,0 +1,1598 @@
...
+                "mod_seq_hash": [1, 1, 1, 2, 2, 2],
+                "mod_seq_charge_hash": [10, 10, 10, 20, 20, 20],
+            }
+        )
+
+        quality_df = pd.DataFrame(
+            {
+                "precursor_idx": [0, 0, 0, 1, 1, 1],
+                "ion": [100, 101, 102, 200, 201, 202],
+                "run1": [0.9, 0.7, 0.3, 0.8, 0.6, 0.4],
```
GeorgWa, line 769: doing the math in my head the third element should be `0.4`

GeorgWa (approved):
LGTM

Image

vuductung (approved):

PR #644, opened 2025-08-05
disable MBR step in default configuration

mschwoer:
Resetting to one-step search as this causes too much confusion (confused people count: 4)


jalew188:
Isn't this a routine step in other engines?

GeorgWa (approved):
For context, this is only in the commandline, not in GUI

PR #647, opened 2025-08-14
Refactor plexscoring

mschwoer:
refactor the scoring postprocessing such that it is accessible as static methods


GeorgWa (approved):
Looking forward to the day when we drop this class.

PR #653, opened 2025-08-21
Refactor calibration iii

mschwoer:
Lint the calibration module with strict rules.

### DIFF
--- a/alphadia/calibration/models.py
+++ b/alphadia/calibration/models.py
@@ -1,3 +1,5 @@
+import logging
+
 import numpy as np
 from sklearn.base import BaseEstimator, RegressorMixin
 from sklearn.preprocessing import PolynomialFeatures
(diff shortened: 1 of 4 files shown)


alphadia/calibration/models.py
```diff
@@ -30,31 +33,34 @@ def __init__(
         n_kernels: int = 6,
         kernel_size: float = 2.0,
         polynomial_degree: int = 2,
-        uniform=False,
+        *,
```
GeorgWa, line 36: why do we need this wildcard?

alphadia/calibration/models.py
```diff
@@ -118,26 +124,23 @@ def _kernel_indices_density(self, x: np.ndarray):
 
         return np.column_stack([start, end]).astype(int)
 
-    def fit(self, x: np.ndarray, y: np.ndarray):
-        """fit the model passed on provided training data.
+    def fit(self, x: np.ndarray, y: np.ndarray) -> "LOESSRegression":  # noqa: C901, PLR0912 Too complex, too many branches
```
GeorgWa, line 127: fun fact, this is one of the oldest pieces of code where I tried to write "software"

alphadia/calibration/property.py
```diff
@@ -390,4 +389,6 @@ def PolynomialRegression(degree: int = 2, include_bias: bool = False) -> Pipelin
 calibration_model_provider = CalibrationModelProvider()
 calibration_model_provider.register_model("LinearRegression", LinearRegression)
 calibration_model_provider.register_model("LOESSRegression", LOESSRegression)
-calibration_model_provider.register_model("PolynomialRegression", PolynomialRegression)
+calibration_model_provider.register_model(
```
GeorgWa, line 392: this model provider was never necessary. Especially now that it's hardcoded we could just drop it.

GeorgWa (approved):
LGTM

alphadia/calibration/property.py
mschwoer: I'd keep it, it's a nice seam and it does not add too much complexity

PR #658, opened 2025-08-26
Ng fixes

mschwoer:
Make the contract of `get_estimators()` compatible with the output writing.
Small refactoring around NG

### DIFF
--- a/alphadia/workflow/base.py
+++ b/alphadia/workflow/base.py
@@ -117,16 +117,15 @@ class WorkflowBase:
             f"Creating DIA data object took: {time.time() - time_start}"
         )  # TODO: debug?
 
-        # alternatively, dia_data.to_ng()
-        time_start = time.time()
-        self._dia_data_ng = (
-            dia_data_to_ng(self._dia_data)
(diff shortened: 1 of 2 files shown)


alphadia/outputtransform/df_builders.py
```diff
@@ -142,12 +142,7 @@ def build_run_stat_df(
             logger.warning(f"Error reading calibration manager for {raw_name}")
 
         prefix = "calibration."
-        for key in [
-            "ms2_median_accuracy",
-            "ms2_median_precision",
-            "ms1_median_accuracy",
-            "ms1_median_precision",
-        ]:
+        for key in calibration_stats:
```
mschwoer, line 145: on second thought it might be better to always have the same columns, but then with nan values, what do you think, @GeorgWa ?

mschwoer: second change: show explicit "NaN" if the column is not given

mschwoer: I chose `NaN` deliberately as the current empty column is confusing when looking at the file from the terminal

PR #660, opened 2025-09-08
Refactor search

mschwoer:
Move all "classic search" related code to the new `search` package.

Only moved code (and adapted imports).


alphadia/calibration/plot.py
```diff
@@ -8,15 +8,82 @@
...
-from alphadia.plotting.utils import density_scatter
+from scipy.stats import gaussian_kde
 
 if TYPE_CHECKING:
     import pandas as pd
 
     from alphadia.calibration.estimator import CalibrationEstimator
 
 
+def _density_scatter(
```
GeorgWa, line 19: I think this is the first time i've seen you making a module larger rather than smaller.
The overall goal is to resolve the mixed bag `alphadia.plotting.utils` ?

GeorgWa (approved):
LGTM

PR #681, opened 2025-09-15
Fix nonconverged estimators

mschwoer:
One column type is not enough to rule them all.


GeorgWa (approved):
putting the state where it belongs

PR #685, opened 2025-09-22
Generify batch plan

mschwoer:
allow the optlock batchplan to grow


alphadia/workflow/optimizers/optimization_lock.py
```diff
@@ -33,7 +33,10 @@ def __init__(self, library: SpecLibFlat, config: Config):
         self._batch_size = config["calibration"]["batch_size"]
 
         self.batch_idx = 0
-        self._set_batch_plan()
+        self.batch_plan = self._get_batch_plan(
+            len(self._elution_group_order),
+            self._batch_size,
+        )
```
GeorgWa, line 39: how do we set `fixed_start_idx`? Don't we need to pass or read this?

PR #694, opened 2025-09-30
Use all cores in distrobuted search

mschwoer:
- pass n cores to alphadia
- set `-e` flag as AlphaDIA now gracefully exits on 0 files


GeorgWa (approved):
LGTM

PR #707, opened 2025-10-26
output refactoring

GeorgWa:
This refators the output creation a bit to be more accessible for introducing the semantic column names.

### DIFF
--- a/alphadia/constants/keys.py
+++ b/alphadia/constants/keys.py
@@ -13,7 +13,7 @@ class ConstantsClass(type):
         ]
 
 
-class StatOutputKeys(metaclass=ConstantsClass):
+class StatOutputCols(metaclass=ConstantsClass):
     """String constants for reading and writing output columns for the `stat` file."""
 
     # optimization
@@ -67,19 +67,19 @@ class InferenceStrategy(metaclass=ConstantsClass):
     HEURISTIC = "heuristic"
 
 
-class SemanticPrecursorKeys(metaclass=ConstantsClass):
+class PrecursorOutputCols(metaclass=ConstantsClass):
     """String constants for accessing the precursor output columns."""
 
     INTENSITY = "precursor.intensity"
 
 
-class SemanticPeptideKeys(metaclass=ConstantsClass):
+class PeptideOutputCols(metaclass=ConstantsClass):
     """String constants for accessing the peptide output columns."""
 
     INTENSITY = "peptide.intensity"
 
 
-class SemanticProteinGroupKeys(metaclass=ConstantsClass):
+class ProteinGroupOutputCols(metaclass=ConstantsClass):
     """String constants for accessing the protein group output columns."""
 
     INTENSITY = "pg.intensity"
--- a/alphadia/outputtransform/utils.py
+++ b/alphadia/outputtransform/utils.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from typing import Literal
 
 import pandas as pd
 from alphabase.peptide import precursor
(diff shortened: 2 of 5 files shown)


alphadia/constants/keys.py
```diff
@@ -57,3 +57,29 @@ class SearchStepFiles(metaclass=ConstantsClass):
...
+
+class InferenceStrategy(metaclass=ConstantsClass):
+    """String constants for protein inference strategies."""
+
+    LIBRARY = "library"
+    MAXIMUM_PARSIMONY = "maximum_parsimony"
+    HEURISTIC = "heuristic"
+
+
+class SemanticPrecursorKeys(metaclass=ConstantsClass):
```
mschwoer: `PrecursorOutputCols`? (also the others)

alphadia/outputtransform/utils.py
```diff
@@ -79,3 +83,207 @@ def write_df(
 
     elif file_format == "tsv":
         df.to_csv(file_path, sep="\t", index=False, float_format="%.6f")
+
+
+def merge_quant_levels_to_psm(
```
mschwoer: (general comment on architecture)
While moving these helper methods to `utils` serves well the purpose of reducing the size of `search_plan_output`, it makes `utils` quite heavy (if you imagine you do the same for the other outputs, this will get crowded)..
I would rather split at the level of
```
        psm_df = self._build_precursor_table(folder_list, save=False)
        self._build_stat_df(folder_list, psm_df=psm_df, save=True)
        self._build_internal_df(folder_list, save=True)
        self._build_lfq_tables(folder_list, psm_df=psm_df, save=True)

        if self.config["general"]["save_mbr_library"]:
            if base_spec_lib is None:
                raise ValueError(
                    "Passing base spectral library is required for MBR library building."
                )
            self._build_mbr_library(base_spec_lib, psm_df=psm_df, save=True)

        if self.config["transfer_library"]["enabled"]:
            self._build_transfer_library(folder_list, save=True)

        if self.config["transfer_learning"]["enabled"]:
            self._build_transfer_model(save=True)
```

to have, e.g. classes called like
```
psm_df = PrecursorTableBuilder(folder_list).build(save=False)
StatDfBuilder(folder_list).build(psm_df=psm_df, save=True)
...
```
Then all methods for building a certain table are in one place in a highly cohesive fashion, yet the coupling between the builders is loose* through defined narrow interfaces.
\* general caveats regarding coupling apply as we still pass around mutable objects

I think it's fine to merge for now (as it's in some way "improving the health of the code base" ;-)) , but if you agree I would pause the current approach and refactor it later.

alphadia/outputtransform/utils.py
```diff
@@ -79,3 +83,207 @@ def write_df(
...
+            "Inference strategy: heuristic. Using maximum parsimony with grouping for protein inference"
+        )
+
+        psm_df = grouping.perform_grouping(
+            psm_df, genes_or_proteins=group_level, group=True
+        )
+
+    else:
+        raise ValueError(
+            f"Unknown inference strategy: {inference_strategy}. Valid options are '{InferenceStrategy.LIBRARY}', '{InferenceStrategy.MAXIMUM_PARSIMONY}' and '{InferenceStrategy.HEURISTIC}'"
```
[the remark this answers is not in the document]
mschwoer: and please add a TODO here to move this check to the config checks (also, add a TODO with a reference to this to `SearchStep._validate_config`)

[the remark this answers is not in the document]
mschwoer: in general: the later we are in the processing, the less change for a raise should be

alphadia/constants/keys.py
mschwoer: hm, a `key` would be rather something that I use to access stuff in a dict, these are quite clearly output `columns` .. please also rename `StatOutputKeys` or add a TODO

PR #749, opened 2025-11-10
adapt to pytest>=9

mschwoer:
Nightly loose tests started failing after pytest upgrade (cf. https://github.com/MannLabs/alphadia/actions/runs/19200448791)


mschwoer:
Did not look into it too much, but given a reason is good practise anyway

PR #810, opened 2026-04-13
Fix tests ii

mschwoer:
Fix tests (detected only by the "full matrix"

@GeorgWa not sure why this did not come up earlier? [b1a13fa](https://github.com/MannLabs/alphadia/pull/810/commits/b1a13fa86c0f5f4761358127cf3cc8737e426f95)

### DIFF
--- a/alphadia/libtransform/prediction.py
+++ b/alphadia/libtransform/prediction.py
@@ -159,11 +159,19 @@ class PeptDeepPrediction(ProcessingStep):
             )
 
         logger.info("Predicting RT, MS2 and mobility")
+
+        process_num = self.mp_process_num
+        if device == "mps" and process_num > 1:
+            logger.info(
+                "Multiprocessing is not supported with MPS device, falling back to single process"
+            )
+            process_num = 1
+
         res = model_mgr.predict_all(
             precursor_df,
             predict_items=["rt", "ms2", "mobility"],
             frag_types=charged_frag_types,
-            process_num=self.mp_process_num,
+            process_num=process_num,
         )
 
         if "fragment_mz_df" in res:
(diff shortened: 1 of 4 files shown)


alphadia/libtransform/prediction.py
```diff
@@ -159,11 +159,19 @@ def forward(self, input: SpecLibBase) -> SpecLibBase:
             )
 
         logger.info("Predicting RT, MS2 and mobility")
+
+        process_num = self.mp_process_num
+        if device == "mps" and process_num > 1:
+            logger.info(
+                "Multiprocessing is not supported with MPS device, falling back to single process"
```
GeorgWa: I’m not aware of this issue tbh. Do we need this?

mschwoer: the tests tell us so? but I am also confused

mschwoer: it was introduced by https://github.com/MannLabs/alphadia/commit/5a8920a18e4581f734d0225f852828ee3a91fd45


```
  reinitialize_ms2_model creates a new pDeepModel() without passing the device — so it defaults to CPU. After the call:
  - ms2_model.device_type is "cpu"
  - peptdeep's guard self.ms2_model.device_type != "cpu" evaluates to False
  - It falls through to the multiprocessing path
  - But the other models (RT, CCS) are still on MPS, and share_memory_() fails on those MPS tensors
```

I patched it here at the two places where `reinitialize_ms2_model` is called .. but peptdeep should preserve the device I guess ..

mschwoer: cf https://github.com/MannLabs/alphapeptdeep/pull/279

GeorgWa: so can we solve this on the peptdeep level?

GeorgWa (approved):
Looks good

PR #814, opened 2026-05-26
add support for .mzML file selection in InputFileSelect component

mschwoer:
Merge input file buttons and add mzML label
image

@GeorgWa : we could include a filter on the extension (such that users can only select e.g. .raw)?


GeorgWa (approved):
Looks good!

PR #817, opened 2026-06-02
Add hdf support

### LINKED ISSUE #311: Can the .hdf files generated by AlphaTims be supported by AlphaDIA?
Hi, I found that AlphaDIA does not support the .hdf from AlphaTims. If that’s the case, then what’s the point of generating the HDF files?

Best,
Jian

mschwoer:
Add support for bruker hdf import. 

Tested for equivalence using `20231218_TIMS03_PaSk_SA_K562_syPASEF_200ng_var_IM0713_S1-E7_1_41539.d`

resolves https://github.com/MannLabs/alphatims/issues/311

### DIFF
--- a/alphadia/raw_data/bruker.py
+++ b/alphadia/raw_data/bruker.py
@@ -4,67 +4,15 @@ import logging
 
 import numba as nb
 import numpy as np
-from alphabase.io.hdf import HDF_File, HDF_Group
 from alpharaw.bruker.timstof import TimsTOFBase
 from alpharaw.utils.pjit import pjit, set_threads
 
 from alphadia.exceptions import NotValidDiaDataError
+from alphadia.raw_data.bruker_hdf import HDF_FILE_EXTENSION, import_data_from_hdf_file
 from alphadia.search.jitclasses.bruker_jit import TimsTOFTransposeJIT
 from alphadia.utils import USE_NUMBA_CACHING
 
 logger = logging.getLogger()
-
-HDF_FILE_EXTENSION = ".hdf"
-
-# Root group name under which alphatims stores a serialized TimsTOF object.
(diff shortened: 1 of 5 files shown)


alphadia/raw_data/bruker.py
```diff
@@ -21,13 +22,18 @@ def __init__(
...
         bruker_d_folder_name: str,
     ):
-        super().__init__(
-            bruker_d_folder_name,
-            mz_estimation_from_frame=1,
-            mobility_estimation_from_frame=1,
-            drop_polarity=True,
-            convert_polarity_to_int=True,
-        )
+        if bruker_d_folder_name.lower().endswith(HDF_FILE_EXTENSION):
```
GeorgWa, line 25: it's unfortunately not that simple, 
HDF can either be an alpharaw hdf or a bruker hdf. I think alpharaw might even be more likely. I think we should try alpharaw import and otherwise go for timstof or we check some of the subpages.

mschwoer: okay, got it .. the bruker hdf functionality is currently in alphatims, and would need to be transferred to alpharaw .. maybe this whole functionality here needs to go to alpharaw as well ..

GeorgWa: a helper method could go to alpharaw. the decision needs to stay in alphadia, as there are (and always will be) still separate raw contaienrs for alpharaw and alphatims.

mschwoer: as discussed via chat: that decision is done here https://github.com/MannLabs/alphadia/pull/817/changes#diff-cf70fea6312f53a56168d895279c3637ebfadc5ea408fdb286ed7bce392e664aR26

GeorgWa (approved):
Looks good!

PR #819, opened 2026-06-16
Peptdeepkontext integraton

mo-sameh:
---

#### To-do list (outside contributers only)
- [ ] I have read and agree to the [MannLabs Contributor License Agreement](https://github.com/MannLabs/.github/blob/main/CLA.md)


alphadia/constants/default.yaml
```diff
@@ -61,6 +66,8 @@ library_prediction:
   # Basic parameters
   # whether to use alphaPeptDeep to predict peptide properties
   enabled: false
+  # whether to use peptdeep_kontext or peptdeep for library prediction
```
mschwoer: `prediction_framework`: `peptdeep` | `peptdeep_kontext`

alphadia/constants/default.yaml
```diff
@@ -124,6 +131,17 @@ library_prediction:
   # minimum probability threshold for including a predicted charge state
   min_charge_probability: 0.1
 
+  # peptdeep_kontext downstream model path; if null, weights are auto-downloaded to ~/.cache/peptdeep_kontext
```
mschwoer, line 135: ideally, this would be in a subsection `peptdeep_kontext`, but I feel we should not nest this yaml too much..

softer measures: mention `peptdeep_kontext` in the comments above `context_path` & `context_indicators`, group it visually `###### peptdeep_kontext ######`

one could also rename to e.g. `peptdeep_kontext_context_indicators`, but thats a bit ugly

alphadia/constants/default.yaml
```diff
@@ -440,6 +458,33 @@ transfer_library:
   # include only fragments with a XIC correlation at least 0.75 of the median for all fragments
   fragment_correlation_ratio: 0.75
 
+context_extraction:
+  # if true, peptdeep_kontext will be used to extract the context and do the library prediction
+  enabled: False
+
+  tto_epochs: 10
```
mschwoer, line 469: add explanatory comments here, and point to the peptdeep_k docs

alphadia/constants/default.yaml
```diff
@@ -14,9 +14,14 @@ general:
 
   # number of threads to use for parallel processing
   thread_count: 10
-  # Enables separate transfer learning step
+
+  # Enables the transfer/adaptation step before the main search (transfer learning or TTO)
+  # TODO: rename to adaptation_step_enabled once backward compat is dropped
```
mschwoer: (nit) isn't the "adaptation" naming the more general that fits both pd and pd_k ? The renaming should be done towards the end of the implementation (in a dedicated PR, which also handles config backwards compat)

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+        """Predict RT and MS2 using peptdeep_kontext context-aware models.
+
+        Mobility is predicted with PeptDeep; RT and MS2 use the pretrained
+        peptdeep_kontext downstream model conditioned on extracted context vectors.
+
+        Parameters
+        ----------
+        use_gpu : bool, optional
+            Use GPU for prediction. Default is True.
+        peptdeep_kontext_model_path : str, optional
```
mschwoer: (nit) `model_path` .. this class has already the context (no pun intended) "peptdeep_kontext"

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+            Path to the pretrained peptdeep_kontext downstream model directory.
+            If not provided, the default bundled model is used.
+        context_path : str, optional
+            Base path to the extracted context file (without ``.json`` extension).
```
mschwoer: so the file is `/bla/blub.json` and I need to pass /bla/blub` ? this is a bit odd .. is there a reason for that?

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+            If not provided, a zero context is used.
+        fragment_types : list[str], optional
+            Fragment types to predict. Default is ["b", "y"].
+        max_fragment_charge : int, optional
+            Maximum fragment charge state to predict. Default is 2.
+        predict_charge : bool, optional
+            Whether to predict charge states using PeptDeep's charge model.
+            Default is False.
+        min_charge_probability : float, optional
+            Minimum probability threshold for including a charge state.
```
mschwoer: an additional sentence explaining what probability this is would make this docstring even better ;-)

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+        if fragment_types is None:
+            fragment_types = ["b", "y"]
+        if indicator_columns is None:
+            indicator_columns = ["constant_context_indicator"]
+
+        super().__init__()
+
+        logger.info(f"Loading peptdeep_kontext model with context path {context_path}")
+
+        self.use_gpu = use_gpu
```
mschwoer: please check which instance variables can be private

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+            ),
+            requested_charged_fragment_types=self.charged_frag_types,
+            dataset_config=PredictionDatasetConfig(
+                feat_extractor="BertaFeatureExtractor",
+                indicator_columns=indicator_columns,
+            ),
+        )
+
+    def validate(self, input: list[str]) -> bool:
+        return True
```
mschwoer, line 101: docstring could give the reason why there's no validation

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def forward(self, input: SpecLibBase) -> SpecLibBase:
+        input.charged_frag_types = self.charged_frag_types
+
+        device = utils.get_torch_device(self.use_gpu)
+
+        # Use PeptDeep for mobility prediction only
```
mschwoer: l.127 tells me it's also used for predicting charge?

alphadia/libtransform/peptdeep_kontext_prediction.py
```diff
@@ -0,0 +1,165 @@
...
+            min_supported = int(charge_range.min())
+            max_supported = int(charge_range.max())
+
+            if "charge" in precursor_df.columns:
+                min_charge = max(min_supported, int(precursor_df["charge"].min()))
+                max_charge = min(max_supported, int(precursor_df["charge"].max()))
+            else:
+                min_charge = min_supported
+                max_charge = max_supported
+
```
mschwoer: don't we have something similar elsewhere? anyway, could be extracted into a small helper method

mschwoer (commented):
first bunch of comments

alphadia/outputtransform/search_plan_output.py
```diff
@@ -172,7 +200,8 @@ def _build_transfer_model(self, save=True):
             ),
         )
         rt_stats = tune_mgr.finetune_rt(transfer_lib.precursor_df)
-        charge_stats = tune_mgr.finetune_charge(transfer_lib.precursor_df)
+        # charge_stats = tune_mgr.finetune_charge(transfer_lib.precursor_df)
+        charge_stats = rt_stats.copy()
```
mschwoer: maybe add comment why the `copy()` is required

alphadia/outputtransform/search_plan_output.py
```diff
@@ -126,6 +128,32 @@ def build(self, folder_list: list[str], base_spec_lib: base.SpecLibBase | None):
         if self.config["transfer_learning"]["enabled"]:
             self._build_transfer_model(save=True)
 
+        if self.config.get("context_extraction", {}).get("enabled", False):
+            self._extract_context()
+    
+    def _extract_context(self) -> None:
+        """Extract the context of the raw files using peptdeep_kontext."""
```
mschwoer, line 140: this can only be done once all raw files have been processed, right? how long does this take roughly?

alphadia/outputtransform/search_plan_output.py
```diff
@@ -246,12 +275,12 @@ def _build_transfer_library(
             f"Built transfer library using {len(folder_list)} folders and {number_of_processes} processes"
         )
         log_stat_df(transfer_library_stat_df(transferAccumulator.consensus_speclibase))
+        transferAccumulator.consensus_speclibase.precursor_df['constant_context_indicator'] = "constant_context"
```
mschwoer: this seems to be important, and is done through a very low-level api .. Ideally, there could be a static method (living in an appropriate pd_k-related class) `set_constant_context_indicator(df) -> None`, with an explanatory docstring .. 
as an alternative, comment on the "why" here

alphadia/transferlearning/context_extraction.py
```diff
@@ -0,0 +1,126 @@
...
+from peptdeep_kontext.datasets.dataset import DatasetConfig
+from peptdeep_kontext.datasets.libloader import LibLoaderConfig
+from peptdeep_kontext.utils.cfg import (
+    download_pretrained_models,
+    get_pretrained_model_config_from_dir,
+)
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_KONTEXT_CACHE = os.path.join(os.path.expanduser("~"), ".cache", "peptdeep_kontext")
```
mschwoer: check out `utils.expand_path()` :-D

alphadia/transferlearning/context_extraction.py
```diff
@@ -0,0 +1,126 @@
...
+        Defaults to ``"context_model_path"``.
+    subdir : str
+        Subdirectory within the downloaded model bundle to return when the
+        explicit path is not set. Defaults to ``"ContextModel"``.
+    """
+    model_path = config.get(path_key)
+    if model_path:
+        return model_path
+
+    version = config.get("context_model_version", "v8")
```
mschwoer: no need for an extra default here, if a key is in `default.yaml`, its value is always available

alphadia/transferlearning/context_extraction.py
```diff
@@ -0,0 +1,126 @@
...
+            tto_dataset_config=DatasetConfig(
+                feat_extractor="BertaFeatureExtractor",
+                indicator_columns=context_indicator_columns,
+            ),
+            tto_lib_loader_config=LibLoaderConfig(
+                root_dirs=[annotated_speclib_path],
+                frag_types=charged_frag_types,
+            ),
+        )
+        self.tto_manager = TTOManager(tto_config)
```
mschwoer: private?

mschwoer (commented):
next  bunch of comments

alphadia/outputtransform/search_plan_output.py
mo-sameh: Nice catch this was actually a debugging artifact

alphadia/outputtransform/search_plan_output.py
mo-sameh: Yes, just like transfer learning. It takes about 4 minutes.

alphadia/outputtransform/search_plan_output.py
mo-sameh: I added a descriptive comment for now. I will address this use case (all from single context)  in pd_k

alphadia/transferlearning/context_extraction.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+from alphadia.utils import expand_path
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_KONTEXT_CACHE = expand_path("~/.cache/peptdeep_kontext")
+
+
+def resolve_context_model_path(
```
mschwoer: from a method called `resolve_..` I would not expect a download .. maybe rename to `prepare_context_model` ?

alphadia/search_plan.py
```diff
@@ -68,21 +67,35 @@ def __init__(
...
+        self._adaptation_method = user_config_general.get(
+            "adaptation_method",
+            default_config_general["adaptation_method"],
         )
         self._mbr_step_enabled = user_config_general.get(
             "mbr_step_enabled", default_config_general["mbr_step_enabled"]
         )
 
+        if self._transfer_step_enabled and self._adaptation_method not in _VALID_ADAPTATION_METHODS:
+            raise CustomError(
```
mschwoer: hm, this is essentially config validation, right? what would happen if l.93-98 is not there, and `_adaptation_method = "blabla"` ?

if the error is "telling enough" (i.e. ugly, but points the user to the problem), I would drop this, for consistency: we do not validate other config values like this, so why start now :-)

if there's no error (but a silent assumption), maybe the downstream component should do the raising?

mschwoer: if you decide to keep it: `UserError`

alphadia/search_plan.py
```diff
@@ -93,17 +106,14 @@ def __init__(
     def _update_paths(self) -> None:
         """Set directories for the different steps.
 
-        If the transfer step is enabled, the quant and library paths for the library step are pointed to the
-            output of the transfer step.
-        If the mbr step is enabled, the quant and library paths for the mbr step are pointed to the output of the
-            library step. Also, the output path for the library step is adjusted to be in a subdirectory of the original output path.
+        If the adaptation step is enabled, the library step output directory is unchanged (it stays
+            at the root output dir), but the adaptation step gets its own subdirectory.
+        If the mbr step is enabled, the library step output is moved to a subdirectory so the root
```
mschwoer: output is not really moved but ends up in its own dir in the first place :-)

alphadia/search_plan.py
```diff
@@ -93,17 +106,14 @@ def __init__(
     def _update_paths(self) -> None:
         """Set directories for the different steps.
 
-        If the transfer step is enabled, the quant and library paths for the library step are pointed to the
-            output of the transfer step.
-        If the mbr step is enabled, the quant and library paths for the mbr step are pointed to the output of the
-            library step. Also, the output path for the library step is adjusted to be in a subdirectory of the original output path.
+        If the adaptation step is enabled, the library step output directory is unchanged (it stays
```
mschwoer: please re-introduce the information on the quant dir

alphadia/search_plan.py
```diff
@@ -170,7 +195,6 @@ def run_plan(self):
         )
 
         if self._mbr_step_enabled:
-            # (outer.sh-steps 4,5)
```
mschwoer, line 173: +1 for boyscouting

alphadia/search_step.py
```diff
@@ -314,27 +316,56 @@ def _parse_modifications(mod_str: str) -> list[str]:
         # 2. Check if properties should be predicted
 
         thread_count = general_config["thread_count"]
-
+        if raw_name is not None and 'constant_context_indicator' not in prediction_config['context_indicators']:
```
mschwoer: did you run pre-commit already? :-)

alphadia/search_step.py
```diff
@@ -314,27 +316,56 @@ def _parse_modifications(mod_str: str) -> list[str]:
...
+            # Combine the cont
```
mschwoer: ?

alphadia/search_step.py
```diff
@@ -437,7 +468,8 @@ def run(
         """
         if self.spectral_library is None:
             logger.progress("Loading spectral library")
-            self.load_library()
+            sample_raw_name = Path(self.raw_path_list[0]).stem if self.raw_path_list else None
```
mschwoer: why are we taking the first only? a docstring on that parameter in `load_library` might answer this ;-)

tests/e2e_tests/prepare_test_data.py
```diff
@@ -85,7 +86,7 @@ def get_test_case(test_case_name: str) -> dict:
 
 if __name__ == "__main__":
     test_case_name = sys.argv[1]
-    target_path = test_case_name
```
mschwoer, line 88: what was the motivation for that change?

alphadia/transferlearning/context_extraction.py
```diff
@@ -0,0 +1,128 @@
...
+    def run(self, save_path: str) -> None:
+        """Extract context from the annotated spectral library and save to disk.
+
+        Parameters
+        ----------
+        save_path : str
+            Base path for the output file. The ``".json"`` extension is appended
+            automatically, producing ``<save_path>.json``.
+        """
+        self._tto_manager.run(f"{save_path}.json")
```
mschwoer: could adding `.json` be part of `run()` such that one just passes `save_path`?

alphadia/outputtransform/search_plan_output.py
```diff
@@ -50,6 +51,7 @@ class SearchPlanOutput:
     LIBRARY_OUTPUT = "speclib.mbr"
     TRANSFER_OUTPUT = "speclib.transfer"
     TRANSFER_MODEL = "peptdeep.transfer"
+    CONTEXT_OUTPUT = "peptdeep_kontext.context"
```
mschwoer: I think the `.json` could be added here .. maybe add two comments
```
# extension omitted as it could be either `.tsv` or `.parquet`
PSM_INPUT = "psm"
..

# extension included 
CONTEXT_OUTPUT = "peptdeep_kontext.context.json"
```

mschwoer (approved):
LGTM, some comments to be addressed

tests/e2e_tests/prepare_test_data.py
mo-sameh: This ensures that when the tests are executed from the `tests` directory, the output of all end-to-end tests is written to a single directory that is already ignored by Git.

alphadia/search_plan.py
mo-sameh: I agree. I removed the implicit assumption and kept the early configuration validation. This allows the pipeline to fail fast, so users don't have to wait for the "FASTA ingestion" and "reading raw files" stages before discovering that the configuration is invalid. It also makes it easier to support additional adaptation methods in the future, such as the combined 'TTO and transfer learning'.

alphadia/constants/default.yaml
GeorgWa: can we write `test time optimization (TTO)` in comment

alphadia/search_plan.py
```diff
@@ -131,35 +144,53 @@ def run_plan(self):
...
                 TRANSFER_STEP_NAME,
             )
 
-            extra_config_from_transfer_step = {
-                ConfigKeys.LIBRARY_PREDICTION: {
-                    ConfigKeys.LIBRARY_PREDICTION.PEPTDEEP_MODEL_PATH: os.path.join(
-                        self._transfer_step_output_dir, SearchPlanOutput.TRANSFER_MODEL
-                    ),
-                    "enabled": True,  # the step following the 'transfer' step needs to have this
+            if self._adaptation_method == "transfer":
```
GeorgWa: can we have a constant object for "transfer" and "tto" ?

GeorgWa (approved):
Looks good!

PR #820, opened 2026-06-24
Use Rust FDR kernels in perform_fdr with exact tie-break handling

GeorgWa:
Adds a Rust-accelerated path for the FDR hot spots and uses it in `perform_fdr`:

- `fdr_q_values` / `fdr_keep_best`: array kernels replacing the pandas multi-key sort + cumsum + groupby.
- `fdr_finalize`: fused sort-free finalization, replacing the `sort -> get_q_values -> keep_best -> get_q_values` chain (up to four O(N log N) sorts) with a per-group arg-min plus a binned histogram of target/decoy counts. Used when fragment competition is not required.
- The reference pandas implementation stays as the fallback and can be forced with `ALPHADIA_RUST_FDR=0` for A/B benchmarking.

`get_q_values` needs an integer tie-break key for the kernel. Precursor FDR breaks ties on `precursor_idx`, but protein FDR breaks ties on `pg`, a string accession. Rather than branching on the column's dtype, non-integer columns are factorized with `sort=True`, which numbers the unique values in sorted order, so ordering by the codes reproduces ordering by the original values and the kernel stays exact for any sortable dtype. Missing values map to the largest key so they sort last, as pandas sorts them.

Requires the Rust kernels from `alphadia-ng` branch `feature/rust-fdr` (`src/fdr.rs`). Until those ship in a released `alphadia-search-rs`, `_RUST_FDR_AVAILABLE` is False, the reference path runs, and `test_fdr_rust_parity.py` skips.


alphadia/fdr/fdr.py
```diff
@@ -326,7 +326,14 @@ def get_q_values(
     if extra_sort_columns is None:
         extra_sort_columns = ["precursor_idx"]
 
-    if _USE_RUST_FDR and len(extra_sort_columns) == 1:
+    # The Rust kernel requires an integer tie-break key. Precursor FDR sorts by
```
mschwoer: these implicit assumptions on sorting will bite us one day :-)

mschwoer (approved):

alphadia/fdr/fdr.py
```diff
@@ -326,7 +326,14 @@ def get_q_values(
...
+    # `precursor_idx` (integer), but protein FDR sorts by the protein-group accession
+    # (a string), so fall back to the reference pandas path for non-integer columns.
+    if (
+        _USE_RUST_FDR
+        and len(extra_sort_columns) == 1
```
mschwoer: Is this fixing a bug? how much of an edge case? or what's the intent?

GeorgWa:
Superseded by #841, which carries the same change plus the exact tie-break handling on a branch cut from a clean `main` (this one was stacked on the unreviewed `feature/rust-fdr`, so merging `main` in inflated the diff to 58 files). Rust side: MannLabs/alphadia-search-rs#130. Closing; `fix/rust-fdr-protein-qval-int64` and `feature/rust-fdr` are being deleted.

PR #821, opened 2026-06-24
Reinitialize and retry collapsed FDR classifier

GeorgWa:
The FDR classifier sometimes gives an almost constant probability (std < `1e-4`). It then cannot separate targets from decoys, all q-values become the global decoy/target ratio, and nothing is filtered. Observed as ~3.5M precursors with one `proba`.

A collapse is usually an unlucky set of start weights, so `perform_fdr` now resets the classifier and fits it again, up to 3 times. It warns if the classifier still collapses. The new `Classifier.reset()` drops the network, so the next fit starts from new random weights.


alphadia/fdr/fdr.py
```diff
@@ -14,15 +14,23 @@
...
+# degenerate to the global decoy/target ratio and no PSMs get filtered.
+_PROBA_COLLAPSE_STD_THRESHOLD = 1e-4
+
+# Number of times to reinitialize and retry a collapsed FDR classifier before giving up.
+_MAX_FDR_CLASSIFIER_REINITS = 3
+
 
 @manage_torch_threads(max_threads=2)
-def perform_fdr(  # noqa: C901, PLR0913, PLR0915 # too complex, Too many statements, too many arguments
+def perform_fdr(  # noqa: C901, PLR0912, PLR0913, PLR0915 # too complex, too many branches, too many statements, too many arguments
```
mschwoer, line 31: one of my favorite methods :-D

alphadia/workflow/managers/fdr_manager.py
```diff
@@ -320,8 +323,13 @@ def load_classifier_store(self, path: None | str = None):
                     )
                     self.classifier_store[classifier_hash].append(classifier)
 
+    @property
+    def classifier_registry(self) -> ClassifierRegistry:
```
mschwoer: does this need to be a public property? why not just inline it?

alphadia/fdr/fdr.py
```diff
@@ -147,6 +160,33 @@ def perform_fdr(  # noqa: C901, PLR0913, PLR0915 # too complex, Too many stateme
...
+    ):
+        n_reinit += 1
+        logger.warning(
+            f"FDR classifier collapsed to a near-constant probability "
+            f"({np.unique(predicted_proba).size} unique value(s) over "
+            f"{len(predicted_proba):,} PSMs); reinitializing from scratch and "
+            f"retrying ({n_reinit}/{_MAX_FDR_CLASSIFIER_REINITS})."
+        )
+        classifier = classifier_registry.get_fresh()
+        classifier.fit(X_train, y_train)
```
mschwoer, line 170: this classifier will not be stored, is this on purpose?

We pass a reference to `classifier`, outside of this `while`, the `fit()` operates on the reference (thus altering the state passed from the outside). If the overwrite `classifier` in l.180, and alter it's state, only the just created fresh copy will be altered, but its scope is local to the method.

alphadia/workflow/managers/fdr_manager.py
```diff
@@ -338,10 +346,8 @@ def get_classifier(self, available_columns: list, version: int = -1) -> Classifi
         """
         classifier_hash = column_hash(available_columns)
         if classifier_hash in self.classifier_store:
-            classifier = self.classifier_store[classifier_hash][version]
-        else:
-            classifier = self.classifier_base
-        return deepcopy(classifier)
+            return deepcopy(self.classifier_store[classifier_hash][version])
+        return self.classifier_registry.get_fresh()
```
mschwoer: `classifier_store` and `classifier_registry` are quite similar concepts, could be integrated into one class .. ( `# TODO` at least)
 
also, from a `registry` I would expect more logic to be exposed (CRUD, versioning, ...) .. maybe call it `classifier_factory` (and `.create()`) ?

alphadia/fdr/fdr.py
mschwoer: Claude's two cents:
- either return the new classifier object (and make sure it is stored in the `classifier_store` or 
- "copying the fresh weights into the original object in place"
```
      fresh = classifier_registry.get_fresh()                                                                                                                                                     
      fresh.fit(X_train, y_train)                                                                                                                                                                 
      # Transplant recovered weights into the caller's object in place                                                                                                             
      classifier.from_state_dict(fresh.to_state_dict())                                                                                                                                           
      predicted_proba = classifier.predict_proba(X)[:, 1]
```

alphadia/fdr/fdr.py
```diff
@@ -14,15 +14,23 @@
...
 
 logger = logging.getLogger()
 
+# Standard deviation below which the FDR classifier's predicted probability is treated
+# as collapsed (near-constant): targets and decoys cannot be separated, so q-values
+# degenerate to the global decoy/target ratio and no PSMs get filtered.
+_PROBA_COLLAPSE_STD_THRESHOLD = 1e-4
+
+# Number of times to reinitialize and retry a collapsed FDR classifier before giving up.
+_MAX_FDR_CLASSIFIER_REINITS = 3
```
mschwoer, line 27: tbh, the whole solution feels more like a patch than a solid fix .. or is the nature of the problem making leaving local minima so unlikely that we are very confident we get to a better solution by just retrying three times?

mschwoer (changes requested):
Requestion changes as this does not work as intended.

alphadia/fdr/fdr.py
```diff
@@ -147,6 +160,33 @@ def perform_fdr(  # noqa: C901, PLR0913, PLR0915 # too complex, Too many stateme
...
+        and n_reinit < _MAX_FDR_CLASSIFIER_REINITS
+    ):
+        n_reinit += 1
+        logger.warning(
+            f"FDR classifier collapsed to a near-constant probability "
+            f"({np.unique(predicted_proba).size} unique value(s) over "
+            f"{len(predicted_proba):,} PSMs); reinitializing from scratch and "
+            f"retrying ({n_reinit}/{_MAX_FDR_CLASSIFIER_REINITS})."
+        )
+        classifier = classifier_registry.get_fresh()
```
mschwoer: add this method to the classifier
```
    def reset(self) -> None:
        """Return the classifier to an unfitted state.

        Drops the trained network so the next `fit` rebuilds it with a new random weight
        initialization. This deliberately does not restore a saved initialization:
        recovering from a collapsed fit requires different starting weights, not the
        original ones.
        """
        self.network = None
        self.optimizer = None
        self._fitted = False
        for values in self.metrics.values():
            values.clear()
```
Important: double-check if this actually serves the purpose of resetting, covers all, and does not have unwanted side effects

and substitute `classifier = classifier_registry.get_fresh()` -> `classifier.reset()`

mschwoer (commented):
update, makes most of previous comments obsolete

alphadia/fdr/fdr.py
GeorgWa: has anybody ever thought about refactoctoring the FDR?

alphadia/fdr/fdr.py
GeorgWa: yes, I really like this approach!

alphadia/fdr/fdr.py
GeorgWa: there was actually a merge conflict with the manmy noqa comments:

alphadia/fdr/fdr.py
GeorgWa: Yes, its a hack. If we run the classifier on the same data, it actually collapses with a 1% probability. something to be addressed separately.

mschwoer (approved):

alphadia/fdr/fdr.py
mschwoer: > has anybody ever thought about refactoctoring the FDR?

there's rumors of a small group of eager scientists planning to tackle that in Q3..

PR #823, opened 2026-06-29
Multiplexting output

mschwoer:
Add per-channel output for multiplexing


GeorgWa (changes requested):
The current approach creates separate files per channel. If we analyse 10 samples with 3 channels this creates the slightly akward situation that we have 3 matrix files with 10 samples each. I think the more natural way is to have a single matrix file with 3x10=30 columns/samples (like DIANN).

PR #825, opened 2026-07-07
Add peptdeep_kontext / TTO support to the GUI

mo-sameh:
Exposes the peptdeep_kontext integration in the GUI so users can select the TTO adaptation method, tune its parameters, and run kontext library prediction — none of which was reachable before.

**Changes**

- PeptideCentric.v1.json — added general.adaptation_method (transfer/tto), library_prediction.prediction_framework as a primary dropdown (+ kontext model/context paths in advanced), and a new context_extraction section for the TTO params.
- WorkflowOverview.js — renamed the node to "Adaptation" with a Transfer / TTO toggle.
- Method.js — conditional section visibility: only the sections relevant to the chosen method show (Transfer → Transfer Library + Transfer Learning; TTO → Transfer Library + Context Extraction).
- multistep.yaml — force-disable context_extraction in the library/mbr steps so it only runs in the adaptation step.

---

#### To-do list (outside contributers only)
- [ ] I have read and agree to the [MannLabs Contributor License Agreement](https://github.com/MannLabs/.github/blob/main/CLA.md)

### DIFF
--- a/gui/src/renderer/components/WorkflowOverview.js
+++ b/gui/src/renderer/components/WorkflowOverview.js
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { useMethod } from '../logic/context'
 import { useMethodDispatch } from '../logic/context';
 import { Masonry } from '@mui/lab';
-import { Box, Typography, Switch, useTheme, List, ListItem, ListItemText, Stack } from '@mui/material';
+import { Box, Typography, Switch, useTheme, List, ListItem, ListItemText, Stack, ToggleButton, ToggleButtonGroup } from '@mui/material';
 
 import ReactFlow, {
     ReactFlowProvider,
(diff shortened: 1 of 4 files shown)


gui/src/renderer/components/WorkflowOverview.js
```diff
@@ -175,31 +175,64 @@ const InputNode = ({ data, id, onEnabledChange }) => {
     );
 };
 
-const TransferLearningNode = ({ data, id, onEnabledChange }) => {
+const TransferLearningNode = ({ data, id, onEnabledChange, onMethodChange }) => {
+    const method = data.adaptationMethod === "tto" ? "tto" : "transfer";
+    const description = method === "tto"
+        ? "Search all files and extract the context with peptdeep_kontext. Will be used for main search."
```
mschwoer: "The result will be used .."

gui/workflows/PeptideCentric.v1.json
```diff
@@ -381,6 +404,20 @@
                         "phospho",
                         "digly"
                     ]
+                },
+                {
+                    "id": "peptdeep_kontext_model_path",
```
mschwoer, line 409: maybe add as prefix "[peptdeep_kontext only]", and add that info also to the description. likewise for the one below, and for "Peeptdeep model path" & ".. type"

gui/workflows/PeptideCentric.v1.json
```diff
@@ -750,6 +794,86 @@
                     ]
                 }
             ]
+        },
+        {
+            "id": "context_extraction",
+            "name": "Context Extraction",
```
mschwoer, line 800: At first I did not get that this changes when clicking "TTO" above .. at least the wording should match ..

ideally (other PR): color clues ..
e.g. the line surrounding "adaptation step" box has a different color than the other steps, the "Transfer library" settings box below has the same. Additionally, the "Transfer"/"TTO" button have different shades of that color, and the respective settings box have that ..

gui/workflows/PeptideCentric.v1.json
```diff
@@ -750,6 +794,86 @@
...
+            "hidden": true,
+            "visible_when": {
+                "general.transfer_step_enabled": true,
+                "general.adaptation_method": "tto"
+            },
+            "parameters": [
+                {
+                    "id": "enabled",
+                    "name": "Enabled for all steps",
```
mschwoer, line 809: (fine for now) 
@GeorgWa this is actually quite confusing ..

mschwoer (approved):

gui/workflows/PeptideCentric.v1.json
mo-sameh: I updated the descriptions and renamed non-obvious parameters to make them clearer. But, using a label like "[PeptDeep only] Peptdeep Model type" makes the name feel a bit too crowded.

gui/workflows/PeptideCentric.v1.json
mo-sameh: I really like the idea of the color clues

PR #826, opened 2026-07-07
Source cycle_fwhm from the NG backend and use it as an FDR feature

mschwoer:
While the rust FWHM was unreliable, `cycle_fwhm` had been hard-coded to `3` and `fwhm_rt` blacklisted from the classifier as a stopgap. With the rust calculation fixed (MannLabs/alphadia-search-rs#122), this puts the real value to work.

- Remove the `cycle_fwhm = 3` hard-code; carry the real per-precursor width from the post-FDR precursors into the quantified output.
- Replace the `fwhm_rt` blacklist with a shared rename map, so `fwhm_rt → cycle_fwhm` stays in sync between the classifier feature-name list and the DataFrame columns — re-enabling it as an FDR feature (matching the python backend, which already uses it).

End-to-end (`basic`, same raw file): precursors **101,614 → 101,674**, proteins **8,457 → 8,504**; reported FWHM median **3.03 s** vs the python reference **2.96 s** (previously a flat `3`).

Depends on MannLabs/alphadia-search-rs#122.


alphadia/workflow/peptidecentric/ng/ng_mapper.py
```diff
@@ -15,6 +15,11 @@
 
 from alphadia.raw_data import DiaData
 
+# NG feature names mapped to their classic equivalents when converting to a classic
+# features DataFrame. Kept in one place so the classifier feature list and the DataFrame
+# columns stay in sync.
+_FEATURE_RENAME_MAP = {"fwhm_rt": "cycle_fwhm"}
```
GeorgWa: I think in this case it would be better to rename the feature in the rust backend and skip the rename

mschwoer: which of the names is better/more descriptive?

alphadia/workflow/peptidecentric/extraction_handler.py
```diff
@@ -630,8 +630,12 @@ def quantify_candidates(
         precursor_df = self.add_columns_from_library(precursor_df, spectral_library)
 
         if precursor_fdr_df is not None:
+            # cycle_fwhm is a scoring feature (not produced by quantification), so it is
+            # carried over from the post-FDR precursors into the quantified output.
             precursor_df = precursor_df.merge(
-                precursor_fdr_df[["precursor_idx", "rank", "qval", "proba"]],
+                precursor_fdr_df[
+                    ["precursor_idx", "rank", "qval", "proba", "cycle_fwhm"]
```
GeorgWa: cycle fwhm can be removed here

alphadia/workflow/peptidecentric/ng/ng_mapper.py
GeorgWa: I like cycle_fwhm

GeorgWa (approved):

alphadia/workflow/peptidecentric/ng/ng_mapper.py
mschwoer: as discussed in person: `fwhm_rt` is the more correct concept

alphadia/workflow/peptidecentric/extraction_handler.py
mschwoer: no, see the comment right above

Issue #132, opened 2024-05-03
Extraction of fragment intensities for transfer learning


GeorgWa:
WHO
Developers

WHAT
Allow for extraction of fragment quantites

WHY
Necessary for MS2 transfer learning

Acceptance Criteria
Ability to set charged frag types, quantities for all fragments, basic filtering

Additional information

PR #842, opened 2026-08-24
Port FragmentCompetition to Rust

GeorgWa:
Replaces the numba/pjit `FragmentCompetition` sweep with a thin wrapper around `alphadia_search_rs.FragmentCompetition`. Drops the now-unused `thread_count` param (Rust uses the crate's global rayon pool).

Depends on MannLabs/alphadia-search-rs#132 — not mergeable until that's released and the requirements pin is bumped.

Test plan:
- [x] `tests/unit_tests/fragcomp`, `tests/unit_tests/fdr` pass locally against a `maturin develop` build of #132
- [ ] full suite once #132 is released and the pin is bumped

### DIFF
--- a/alphadia/fragcomp/fragcomp.py
+++ b/alphadia/fragcomp/fragcomp.py
@@ -1,9 +1,8 @@
 """The fragment competition module contains functionality to maintain the exclusive assignment of signal to identifications.
 
-Thin Python wrapper around the Rust fragment-competition kernel
-(`alphadia_search_rs.FragmentCompetition`). The numeric sweep (per-DIA-window
-fragment-overlap comparison) lives in Rust; this wrapper only prepares the
-candidate/fragment index bookkeeping and dataframe I/O.
+The numeric work — DIA-window assignment, priority ranking and the fragment-overlap
+sweep — lives in Rust (`alphadia_search_rs.FragmentCompetition`). This module only
+prepares the candidate/fragment index bookkeeping around it.
 """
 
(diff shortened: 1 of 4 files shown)


mschwoer (approved):

alphadia/fragcomp/fragcomp.py
```diff
@@ -1,300 +1,88 @@
...
-    int
-        The number of overlapping fragments.
-
-    """
-    frag_mz_1 = frag_mz_1.reshape(-1, 1)
-    frag_mz_2 = frag_mz_2.reshape(1, -1)
-    delta_mz = np.abs(frag_mz_1 - frag_mz_2)
-    ppm_delta_mz = delta_mz / frag_mz_1 * 1e6
-    return np.sum(ppm_delta_mz < mass_tol_ppm)
+DEFAULT_RT_TOLERANCE_SECONDS = 3.0
```
mschwoer, line 20: what motivates those values? shouln't they be optimized/configurable?

GeorgWa: this should be the fwhm of each peak :D

mschwoer: like, `fwhm_rt` from here? https://github.com/MannLabs/alphadia-search-rs/pull/122
but currently we don't pass it?

mschwoer: AX-1532

PR #846, opened 2026-08-30
Fix ZeroDivisionError when an optimization batch extracts nothing

GeorgWa:
A search stops with a `ZeroDivisionError` when an optimization batch extracts nothing. The failure occurs in the code that must recover from a batch with too few precursors.

Three optimizers measure the same quantity. Each one calculated it separately, and none of them handled a batch with no data. This change gives the measurement its own class. The class holds the name, the calculation, and the rule for a batch with no data. All optimizers now use one rule, and a new optimizer gets the rule with it.

A round with no data is left out of the optimization history. The alternative is to record a zero. But zero is also a valid measurement, and it would corrupt the convergence test without an error message.

Optimization keeps the current search parameter when a round gives no data. It also completes normally when no round gives data at all.

A batch with data behaves as before.

### DIFF
--- a/alphadia/workflow/optimizers/automatic.py
+++ b/alphadia/workflow/optimizers/automatic.py
@@ -42,14 +42,6 @@ class AutomaticOptimizer(BaseOptimizer, ABC):
         See base class for other parameters.
 
         """
-        # Fail here rather than minutes into the first search: without the abstract
-        # _get_feature_value, a subclass that forgets to declare its feature would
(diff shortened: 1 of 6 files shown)


GeorgWa:
Superseded by a three-PR stack built from a clean main. The None-return design in this branch left several edge cases open in review, so the redesign excludes them at the handler boundary instead.

mschwoer:
could you also delete the stale branch(es) @GeorgWa  ?

PR #851, opened 2026-09-15
Separate directLFQ ion table preparation from intensity estimation

GeorgWa:
The next PR replaces directLFQ's median of fragment ratios at the precursor level with a different estimator that still needs directLFQ's normalised log2 ion table. `direct_lfq` had table preparation, sample normalisation and estimation in one block, so this extracts the first two into their own methods and dedupes the fragment metadata column list shared with `filter_frag_df`.

No behaviour change. First of two stacked PRs.


alphadia/outputtransform/quantification/quant_builder.py
```diff
@@ -16,6 +16,15 @@
 
 logger = logging.getLogger()
 
+# Columns of the accumulated fragment matrices that hold metadata rather than run intensities
```
mschwoer, line 19: this solves the bug that new metadata columns are treated as intensities, right?
It's a step -- but I am questioning the direction: 
if one forgets to add new metadata columns here, we have the same problem.

Aren't the raw file/run names known in this context? could we not opt-in on columns instead of opt-out?

mschwoer: edit: doesn't solve the bug, the bug is all about forgetting to add new columns .. but then fair enough, my comment above is not in scope of this change

mschwoer (approved):
+1 for separate upfront refactoring

alphadia/outputtransform/quantification/quant_builder.py
GeorgWa: I think the only way to handlle this properly is to move away from pandas dataframes as interfaces.

PR #852, opened 2026-09-15
Correlation-weighted fragment sum as the foundation of quantification

GeorgWa:
directLFQ's median of fragment ratios gives every fragment an equal vote, so interfered, poorly correlating fragments compress true fold changes. In the research harness on the ProteoBench Astral bulk data this rollup lowered the species-averaged median |eps| from 0.192 to 0.158 at identical identifications.

Precursor quantities are now the sum of normalised fragment intensities, each fragment weighted by the fourth power of its mean cross-run correlation. One weight per fragment, shared by all runs, so ratios between runs are preserved. Peptides and protein groups are estimated by directLFQ from these precursors, so sample normalisation happens once, on the fragment table.

Output change: a precursor is reported in every run with an observed fragment, so `min_nonnan` now applies to the peptide and protein group levels only.

Still to do before merge: the fixed-identification re-roll on ProteoBench data. The precursor-based protein group rollup was not part of the research numbers above.

Stacked on #851.


alphadia/outputtransform/quantification/quant_builder.py
```diff
@@ -25,6 +25,13 @@
     "mod_seq_charge_hash",
 ]
 
+# Fragment weights are the mean cross-run correlation raised to this power. Higher values let
+# well-correlating fragments dominate: 2 favours precision, 6-8 favour the accuracy of spiked-in
```
mschwoer: "favour the accuracy of spiked-in ratios"?

alphadia/outputtransform/quantification/quant_builder.py
```diff
@@ -223,6 +266,56 @@ def direct_lfq(
...
+        self,
+        intensity_df: pd.DataFrame,
+        config: Config,
+        ion_quality: pd.Series,
+    ) -> pd.DataFrame:
+        """Roll fragments up to precursors by a correlation-weighted sum of their normalized intensities.
+
+        Every fragment gets a single weight shared by all runs, so ratios between runs
+        are preserved. Raising the cross-run correlation to a steep power lets the
+        well-behaved fragments dominate, where a median of fragment ratios gives every
```
mschwoer: "well-behaved " -> please elaborate

PR #853, opened 2026-09-16
enforce int64 signature on ion hash kernel

mschwoer:
Fix potential collision in `_ion_hash` due to numba casting, add explicit signature to detect this issue for future callers.

Sloppidislop:

## Problem

`_ion_hash` is a numba kernel that shifts `number`, `type`, `charge`, `loss_type` into the
upper bits of `precursor_idx`. `prepare_df` passed `df["precursor_idx"].values`, which is
unsigned. Mixing an unsigned array with the signed shifted terms makes numba promote the
result to `float64`, whose 53-bit mantissa silently rounds away the low bits of large hashes.
Colliding hashes then merge fragments of different precursors in the LFQ accumulation.

## Change
- `quantification/quant_builder.py:127` — cast to `int64` before hashing:
   ` df["precursor_idx"].values.astype(np.int64)`


GeorgWa (approved):
looks good!

PR #854, opened 2026-09-16
Reuse quant from

mschwoer:
Add the option to reuse quants from multiple distributed previous single-file runs.

Deliberately mutually exclusive with `reuse_quant` as the combination is unusual and unwanted effects are likely.

reimplementation of this previous spike: https://github.com/MannLabs/alphadia/compare/main...reuse-quant-distributed-search


GeorgWa (commented):
I think it's seantically a bit counterintuitive that `reuse_quant` and `reuse_quant_from `are exclusive. 

From a user perspective I expect `reuse_quant` : True to reuse quant from default folder and then having an ovverride for the default folder in `quant_folders=[]`

GeorgWa (approved):
Looks good implementation wise. I think it would be cleaner if reuse quant and quant path would work together.

mschwoer:
> I think it's seantically a bit counterintuitive that `reuse_quant` and `reuse_quant_from `are exclusive.
> 
> From a user perspective I expect `reuse_quant` : True to reuse quant from default folder and then having an ovverride for the default folder in `quant_folders=[]`

True that, the part I was missing is that the `reuse_quant` flag could trigger the logic, and `reuse_quant_from` only provides folders ..
changed

> I think it would be cleaner if reuse quant and quant path would work together.
are you referring to the above comment, or is this something new?

alphadia/constants/default.yaml
```diff
@@ -25,6 +25,10 @@ general:
   reuse_calibration: false
   # whether to reuse previously calculated quantification data
   reuse_quant: false
+  # quant directories (e.g. `<previous_run>/quant`) of other runs to reuse quantification data from
+  # instead of the quant directory of the current run, requires `reuse_quant`.
+  # These directories are only read from, never written to.
+  reuse_quant_from: []
```
vbrennsteiner: [nit] I think this argument states more than it needs to, in the end the fact _that_ we "reuse_quant" is already set above, so this could be "quant_folders" as @GeorgWa suggested?

GeorgWa:
I like it with the changed interface!

mschwoer:
@vbrennsteiner we could think about adapting distributed_quant now (cf. draft here https://github.com/MannLabs/alphadia/compare/main...reuse-quant-distributed-search) .. basically not have step `n` write to a directory of step `n+1` but step `n+1` reading the output of step `n`..

PR #855, opened 2026-09-16
read only needed fragment columns and merge on ion hash alone

mschwoer:
From an earlier timing analysis: some low-hanging fruits to reduce memory and runtime.

- use only required columns (a common pandas pattern values by many advanced proteomics software specialists)
- drop redundant second merge key

reimplementation of this previous spike: https://github.com/MannLabs/alphadia/compare/main...reuse-quant-distributed-search


GeorgWa (approved):
Looks good to me!

mschwoer:
back to draft .. will merge later to not interfere with ongoing changes

PR #857, opened 2026-09-16
Tweaks for reusing quant

mschwoer:
Fixing some minor issue that prevented my re-using older versions.

I think it's fine to just warn in case older managers are being reloaded. Unless we do subtle changes of units (e.g. min-> sec) there, any errors should surface downstream


alphadia/workflow/managers/base.py
```diff
@@ -94,17 +96,20 @@ def load(self):
...
-                    self.is_loaded_from_file = True
-                    self.reporter.log_string(
-                        f"Loaded {self.__class__.__name__} from {self.path}"
-                    )
-                else:
-                    self.reporter.log_string(
-                        f"Version mismatch while loading {self.__class__}: {loaded_state._version} != {self._version}. Will not load.",
-                        verbosity="warning",
-                    )
+            # TODO: checking against AlphaDIA version is too strict, ideally each manager has its own (strictly semantic) version
```
GeorgWa, line 99: Wouldn't this be the best, to check for semver compatability?

GeorgWa (approved):
Looks good!

alphadia/workflow/managers/base.py
mschwoer: even if we used proper semver for alphadia, this would mean you make this part of the public API. Which can be done of course, on the other hand it's "just" a convenience feature to save compute.
I was more thinking that each Manager gets its own semantic version, and the checks are done individually .. but probably over-eng'd

PR #858, opened 2026-09-16
Refactor LFQ method decision

GeorgWa:
Preparation for the correlation-weighted precursor rollup in #860. No change in what is computed.

- `build` decides once whether directLFQ or QuantSelect is used, instead of per level.
- The per-level `normalization_method` field is removed; nothing read it.
- `direct_lfq` is split into ion table preparation, optional sample normalisation and estimation.
- `direct_lfq` returns an empty frame instead of failing inside directLFQ when nothing was observed.
- The fragment matrix column names are constants shared by the filter and the ion table preparation.
- The correlation matrix is called `correlation_df` instead of `quality_df`.
- `save_fragment_matrices` and the `save_fragments` flag are removed; nothing called them.


alphadia/outputtransform/quantification/quant_output_builder.py
```diff
@@ -99,27 +99,31 @@ def build(
...
-                continue
+        requested_configs = [c for c in quantlevel_configs if c.should_process]
 
-            logger.info(
-                f"Performing label free quantification on the {quantlevel_config.level_name} level"
+        if (
+            self.config["search_output"]["normalization_method"]
+            == NormalizationMethods.QUANTSELECT
+        ):
+            level_dfs = self._quantify_with_quantselect(
```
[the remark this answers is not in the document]
GeorgWa: While this would be good, it's beyond the scope of this refactoring as it touches the quantselect implementation.

alphadia/outputtransform/quantification/quant_builder.py
```diff
@@ -212,11 +210,12 @@ def direct_lfq(
...
         """
-        logger.info(
-            f"Performing label-free quantification with {lfq_config.normalization_method} normalization"
-        )
+        logger.info("Performing label-free quantification with directLFQ")
 
         lfq_df = self._prepare_ion_table(intensity_df, lfq_config.quant_level)
+        # directLFQ's normalization divides by the number of ions
+        if lfq_df.empty:
+            return pd.DataFrame(columns=[lfq_config.quant_level])
```
mschwoer: is this enough columns in this case?

alphadia/outputtransform/quantification/quant_output_builder.py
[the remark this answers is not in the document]
mschwoer: add as a TODO?

alphadia/outputtransform/quantification/quant_output_builder.py
```diff
@@ -151,12 +155,6 @@ def _create_quant_level_configs(self) -> list[LFQOutputConfig]:
                     "charge",
                 ],
                 should_process=self.config["search_output"]["precursor_level_lfq"],
-                save_fragments=self.config["search_output"][
-                    "save_fragment_quant_matrix"
-                ],
-                normalization_method=self.config["search_output"][
-                    "normalization_method"
```
mschwoer, line 158: could those keys then be removed from the default.yaml? (check the non-breaking mechanism in config.py)

alphadia/outputtransform/quantification/quant_builder.py
```diff
@@ -212,11 +210,12 @@ def direct_lfq(
         pd.DataFrame
             Protein/peptide quantification results with columns: group_column, run1, run2, ...
         """
-        logger.info(
-            f"Performing label-free quantification with {lfq_config.normalization_method} normalization"
-        )
+        logger.info("Performing label-free quantification with directLFQ")
 
         lfq_df = self._prepare_ion_table(intensity_df, lfq_config.quant_level)
+        # directLFQ's normalization divides by the number of ions
```
mschwoer: is this comment at the right place?

alphadia/outputtransform/quantification/quant_output_builder.py
```diff
@@ -320,53 +335,3 @@ def save_results(
                 os.path.join(output_folder, f"{config.level_name}.matrix"),
                 file_format=file_format,
             )
-
-    def save_fragment_matrices(
```
mschwoer, line 324: wasn't this required by some downstream processing?

mschwoer (approved):

WassimG (approved):

alphadia/outputtransform/quantification/quant_output_builder.py
GeorgWa: yes, for Tungs quantselect imlementation before it was integrated. can be safely dropped.

alphadia/outputtransform/quantification/quant_output_builder.py
GeorgWa: good point, maybe it's safer to keep it

alphadia/outputtransform/quantification/quant_builder.py
GeorgWa: an empty frame only has the group column, this is intended.

alphadia/outputtransform/quantification/quant_output_builder.py
mschwoer: if they are no longer references, it's safe to remove I guess :-D

alphadia/outputtransform/quantification/quant_builder.py
mschwoer: so the methods somehow needs to know what the caller requires in terms of columns .. we could also just return `None` here and let the caller do the handling
(no action required, it's just some patterns that we are regularly using that I am flagging ;-))

alphadia/outputtransform/quantification/quant_output_builder.py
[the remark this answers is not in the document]
GeorgWa: Tracked as a TODO at the dispatch (42148465). The build-level test touches the QuantSelect code and stays out of this refactor.

PR #859, opened 2026-09-16
Correlation-weighted fragment sum as the foundation of quantification

GeorgWa:
directLFQ's median of fragment ratios gives every fragment an equal vote, so interfered, poorly correlating fragments compress true fold changes. In the research harness on the ProteoBench Astral bulk data this rollup lowered the species-averaged median |eps| from 0.192 to 0.158 at identical identifications.

Precursor quantities are now the sum of the directLFQ-normalised fragment intensities, each fragment weighted by the fourth power of its mean cross-run correlation relative to the best fragment of its precursor. One weight per fragment for all runs keeps the ratios between runs. Peptides and protein groups are estimated by directLFQ from these precursors, so normalisation happens once.

Consequences for existing knobs: `min_nonnan` now counts precursors per sample and defaults to 1, and the fragment filter runs per precursor, which the YAML, GUI and docs now say. A precursor is reported in every run with an observed fragment.

Still to do before merge: the fixed-identification re-roll on ProteoBench data. The precursor-based protein group rollup was not part of the research numbers above.

Stacked on #858. Supersedes #852.


GeorgWa:
Superseded by #860 (feature) and #861 (cleanup), split for review.

PR #860, opened 2026-09-16
Correlation-weighted fragment sum as the foundation of quantification

GeorgWa:
Precursor quantities become the foundation of quantification. Stacked on #858; #861 removes the fragment filter on top.

- A precursor is the sum of its normalised fragment intensities, each fragment weighted by the fourth power of its mean cross-run correlation relative to the precursor's best fragment.
- One weight per fragment for all runs, so ratios between runs are preserved. directLFQ's median of fragment ratios gave interfered fragments an equal vote and compressed fold changes.
- Peptides and protein groups are estimated by directLFQ from the precursor quantities, so sample normalisation runs once, on the fragment table.
- A precursor listed under several protein groups counts for each of them.
- `min_nonnan` now counts precursors per sample and defaults to 1.
- A precursor is reported in every run with an observed fragment.
- In the research harness on ProteoBench Astral bulk data this rollup lowered the species-averaged median |eps| from 0.192 to 0.158 at identical identifications. The precursor-based protein group rollup still needs the fixed-identification re-roll before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #861, opened 2026-09-16
Drop the fragment filter and its two config keys

GeorgWa:
Removes the fragment filter that the new rollup made redundant. Stacked on #860.

- The sum runs on every quantified fragment; `filter_frag_df` and its tests are gone.
- `min_k_fragments` and `min_correlation` are removed from the defaults and the GUI. Grouped per precursor, the top-12 kept every fragment the search scores, so the correlation cut never removed anything.
- Both keys are listed in `TOLERATED_KEYS`: a user config that still sets them gets the usual removal warning instead of an error, until the next major version.


mschwoer (approved):
LGTM

PR #863, opened 2026-09-16
Fail fast when the search extracts no candidate

GeorgWa:
## Summary

When the optimization lock runs out of batches before reaching its target, `search_parameter_optimization` measures the last round and proceeds with the best parameters available. If the whole raw file yielded zero candidates, `total_elution_groups` is 0 and the precursor proportion feature divided by zero deep inside `proceed_with_insufficient_precursors`.

This PR raises `NoOptimizationLockTargetError` at the handler boundary instead, so that

- the raw file lands in `raw_files_with_errors` (or aborts under `fail_fast`) with a clear message instead of a `ZeroDivisionError`, and
- no optimizer ever sees an optimization lock without elution groups, which lets the feature computations stay total functions.

The recovery path for files with candidates but too few precursors is unchanged.

## Tests

- `test_search_parameter_optimization_raises_without_candidates`
- `test_search_parameter_optimization_recovers_with_insufficient_precursors`

## Stack

Replaces #846, split for review. Merge order: #863, #864, #865.


alphadia/workflow/peptidecentric/optimization_handler.py
```diff
@@ -275,6 +276,12 @@ def search_parameter_optimization(self):
                 if not self._optlock.has_target_num_precursors:
                     log_string("Target number of precursors not reached yet.")
                     if not self._optlock.batches_remaining():
+                        if self._optlock.total_elution_groups == 0:
+                            # The recovery path below measures the last round. Without
```
mschwoer, line 280: recovery path? measure?

mschwoer (approved):

alphadia/workflow/peptidecentric/optimization_handler.py
GeorgWa: sorry for slop comments :/

PR #864, opened 2026-09-16
Declare optimizer identity as class attributes

GeorgWa:
## Summary

Each optimizer subclass only existed to set four constants (`parameter_name`, `_estimator_group_name`, `_estimator_name`, `_feature_name`) and forward seven constructor arguments to its parent. This PR declares the constants as class attributes and removes the eight near-identical `__init__` methods in `automatic.py` and `targeted.py`.

Pure refactor with no behaviour change. The existing tests construct every subclass and assert its parameter name.

## Stack

Replaces #846, split for review. Merge order: #863, #864, #865.


mschwoer (approved):
LGTM

PR #865, opened 2026-09-16
Give the optimizer feature a home of its own

GeorgWa:
## Summary

The feature an automatic optimizer maximises was a history column name (`_feature_name`) in one place and a computation (`_get_feature_value`) in another, repeated per subclass. This PR introduces `OptimizationFeature` in `workflow/optimizers/features.py`: a stateless class that keeps the column name and the computation together. Optimizers reference the feature class itself.

Design points:

- `measure` is a total function. `PrecursorProportionDetected` relies on the handler precondition from #863 that the lock holds at least one elution group. `MeanIsotopeIntensityCorrelation` is NaN without precursors, and pandas ignores NaN when it searches for the maximum, so such a round is never the optimum.
- `_feature` is an abstract property on `AutomaticOptimizer`, so a subclass that declares no feature fails at construction, exactly as the abstract `_get_feature_value` did.
- No `None` return and no empty-history guards. The existing check in `_find_index_of_optimum` is the single remaining assertion, and it is unreachable from `step()` and `skip()`.

Stacked on #864 (base branch `fix/declarative-optimizers`). Merge order: #863, #864, then this PR.

## Tests

- `test_optimization_features.py`: ratio, zero for empty precursors, mean, NaN without precursors
- `test_automatic_optimizer_requires_a_feature`
- `test_automatic_ms1_optimizer_does_not_pick_a_round_without_precursors`

## Known gap, not addressed here

The rust backend emits no `isotope_intensity_correlation` column, so `AutomaticMS1Optimizer` (`target_ms1_tolerance <= 0`) raises `KeyError` at its first step under the default backend. This predates this stack.


alphadia/workflow/optimizers/features.py
```diff
@@ -0,0 +1,77 @@
...
+
+from abc import ABC, abstractmethod
+
+import pandas as pd
+
+from alphadia.workflow.optimizers.optimization_lock import OptimizationLock
+
+
+class OptimizationFeature(ABC):
+    """The value that an AutomaticOptimizer makes as large as possible.
```
mschwoer, line 11: "The feature that the optimizer makes as large as possible", "The value that an AutomaticOptimizer makes as large as possible." ..
sounds like we are lacking a clear name for that concept..
`target_feature`?
meh ..

mschwoer (commented):
nice one!

alphadia/workflow/optimizers/features.py
GeorgWa: yes, we are basically introducing the name and concept here and we call it `OptimizationFeature` now. I think we had a brief discussion about this in person and came up with this.

GeorgWa: we also discuess calling it metric or similar

mschwoer (approved):

PR #867, opened 2026-09-16
Correct RT-dependent intensity drift before label-free quantification

GeorgWa:
Plan 03 of the quantification work: remove each run's intensity drift along the gradient before directLFQ. Paired with the GUI-only #868; merge this one first, since the config rejects keys that are not in the defaults.

**What it does**
- New `IntensityDriftCorrector` in `outputtransform/quantification/intensity_drift_correction.py`. Per run it takes the log2 deviation from the across-run ion median over well-correlating fragments (mean correlation ≥ 0.8, ≥ 3 observations), bins it in 30 s RT bins merged to ≥ 100 ions, reads the smoothed-histogram mode per bin, smooths the bin curve over 3 points and interpolates it to every ion.
- The curve is re-centred to zero mean over the run's ions, so only the shape along RT is removed and the run level stays with the directLFQ normalisation. Runs with fewer than 500 usable ions are left unchanged.
- Applied once to the accumulated fragment table in `QuantOutputBuilder._quantify_with_directlfq`, before the per-level fragment filter, so the precursor, peptide and protein levels see the same corrected intensities. QuantSelect is untouched.
- Ions with an intensity but no `rt_observed` in a run are placed at their mean RT over the other runs. If the psm table lacks `rt_observed` or `run`, the correction is skipped with a warning.
- Single switch `search_output.intensity_drift_correction: True`; the estimator parameters are constants on the class. The sub-sampled curve is logged per run at INFO.

**Tests**
- 12 unit tests on synthetic 6-run data: drift removal for a single run and for a whole condition, shape-only (every run mean unchanged), no spurious correction without drift, curve built from well-correlating ions only, zeros and metadata preserved, inputs not modified, missing-column skip, too-few-ions skip, RT fallback, bin merging.
- 2 builder tests: the correction runs once before filtering when enabled and not at all when disabled.

**Reviewer notes**
- On synthetic data `mode` removes drift less completely than `median` in a balanced 3-vs-3 design (residual slope 0.03 vs 0.01 log2 per 1000 s from an initial 0.20). The across-run median sits between the conditions and halves a run's own deviation whenever the run is one of the two middle values, which skews the distribution the mode reads. `mode` is kept as in the plan; `IntensityDriftStatistic.MEDIAN` is implemented and one constant away.
- The plan's acceptance criteria (bulk fixed-ID re-roll, low-input eps by RT sextile) still need an e2e run.


GeorgWa:
Superseded by #870, which carries the same feature with the estimator moved to the Rust backend (MannLabs/alphadia-search-rs#142) and the branch rebuilt on current main. The numpy implementation reviewed here is reproduced to 2e-15 log2 by the Rust one. Closing in favour of #870; the branch is kept for reference.

